Issue #66, opened 2026-05-21
MCP akb_search omits user_id / vault read check, risking unauthorized search results


MackDing:
## Summary
The MCP `akb_search` handler currently calls `SearchService.search()` without passing the authenticated `uid`, and it also does not perform a vault-level read check when a `vault` argument is supplied.

## Evidence
In `backend/mcp_server/server.py`:

```py
@_h("akb_search")
async def _handle_search(args: dict, uid: str, user: _MCPUser) -> dict:
    result = await search_service.search(
        query=args["query"],
        vault=args.get("vault"),
        collection=args.get("collection"),
        doc_type=args.get("type"),
        tags=args.get("tags"),
        limit=args.get("limit", 10),
    )
    return result.model_dump()
```

But in `backend/app/services/search_service.py`, ACL prefiltering is only applied when `user_id` is provided:

```py
has_filters = any([vault, collection, doc_type, tags, user_id])
...
# Always pre-filter when user_id is provided so we never leak
# documents from vaults the user can't read.
```

So MCP search requests can bypass the access-control path that the REST layer uses.

## Risk
A caller using MCP may be able to retrieve hits from vaults they should not be able to read, especially for cross-vault or loosely filtered searches. Even when a vault name is provided, the handler does not currently reject unauthorized access up front.

## Expected fix
- pass `user_id=uid` into `search_service.search(...)` in the MCP handler
- when `vault` is supplied, also enforce `check_vault_access(uid, vault, required_role="reader")` for a fail-closed path aligned with other MCP handlers like `akb_browse` and `akb_get`
- add/extend an e2e regression test covering unauthorized MCP search access

I have a patch ready for this and can open a PR.

PR #67, opened 2026-05-21
fix: enforce ACLs for MCP search

MackDing:
## Summary
- pass the authenticated `uid` into `SearchService.search()` from the MCP `akb_search` handler
- fail closed with `check_vault_access(..., required_role="reader")` when a vault is explicitly targeted
- extend the MCP e2e flow to assert search is denied again after revoke

## Why
Issue #66: the MCP path was bypassing the ACL-aware `user_id` prefilter in `SearchService.search()`, which created a risk of unauthorized search results.

## Verification
- `python3 -m py_compile backend/mcp_server/server.py`
- `bash -n backend/tests/test_mcp_e2e.sh`

### DIFF
--- a/backend/mcp_server/server.py
+++ b/backend/mcp_server/server.py
@@ -310,6 +310,8 @@ async def _handle_browse(args: dict, uid: str, user: _MCPUser) -> dict:
 
 @_h("akb_search")
 async def _handle_search(args: dict, uid: str, user: _MCPUser) -> dict:
+    if args.get("vault"):
+        await check_vault_access(uid, args["vault"], required_role="reader")
     result = await search_service.search(
         query=args["query"],
         vault=args.get("vault"),
@@ -317,6 +319,7 @@ async def _handle_search(args: dict, uid: str, user: _MCPUser) -> dict:
         doc_type=args.get("type"),
         tags=args.get("tags"),
         limit=args.get("limit", 10),
+        user_id=uid,
     )
     return result.model_dump()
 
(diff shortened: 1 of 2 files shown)


byongjohnhan-oss:
Merged as f04eb18. Thanks for the second contribution @MackDing — a security fix this time, not just perf. The repro path was empirical:

**Pre-fix** (main at f8f7a71): a non-member calling `akb_search` with no vault arg returned 5 hits including the private-vault secret doc — the missing `user_id=uid` caused the ACL prefilter inside `SearchService.search()` to no-op.

**Post-fix**:
- `akb_search({query: "..."})` → 0 hits (ACL prefilter scopes to caller's vaults)
- `akb_search({query: "...", vault: "<other-vault>"})` → `{"error": "Requires 'reader' role on vault '...'"}`

Both paths properly closed. CI green, regression suites pass.

Two follow-ups your repro surfaced (not blockers, will track separately):
1. The service-layer `if vault is None and user_id is None: raise ValidationError` guard in `search_service.search` didn't actually fire on the pre-fix path — Bob reached the SQL stage and got results. Worth a separate look at the guard's reachability.
2. Audited the other MCP handlers for the same propagation pattern. One more gap found: `akb_help(topic="vault-skill", vault=...)` calls `doc_service.get` without `check_vault_access` and can leak a private vault's vault-skill content. Filing as a follow-up fix.

Deploying to prod now. Appreciate the careful review.

byongjohnhan-oss:
Correction on follow-up #2 above — after a closer look + maintainer review, the `akb_help(topic="vault-skill")` lack of `check_vault_access` is **intentional**, not a bug. The vault-skill doc is meta about how to write into the vault (conventions, doc types, tag taxonomy) rather than vault data; treating it as discoverable-by-any-authenticated-user matches the documented agent flow ("`akb_help(topic="vault-skill", vault=...)` before writing"). It's the same model as a public README for a private repo.

So scratch that one — no fix needed there. Follow-up #1 (service-layer ValidationError guard reachability) still stands as a separate look.

PR #100, opened 2026-06-02
chore: restore CI green — fix 22 mypy errors (check) + _FakeDocRepo for conn= (backend-pytest)

kwoo24-oss:
## Problem

`scripts/check.sh` runs `mypy app/ mcp_server/` as a CI gate, but it had been **red on 22 pre-existing errors** across 6 files — long-standing `Optional`-strictness debt, not introduced by any recent change. (Confirmed: the same 22 errors are present back through the pre-existing history; they predate the recent deadlock fix.)

## Fix

All 22 fixed **properly** — type guards and precise annotations, **no `# type: ignore` suppressions**:

| File | Fix |
|---|---|
| `repositories/document_repo.py` | Move the `like_escape` import from the class body to module top (`Unsupported class scoped import`). |
| `services/kg_service.py` | Fail-closed guard when a linkable URI has no identifier (`str\|None` → `str` for the advisory-lock + existence checks); skip identifier-less URIs in the batch name resolver. |
| `services/resource_integrity.py` | `ResourceHashRepairReport.errors` is a non-Optional `list[str]` (`field(default_factory=list)`); a `FileHashProjection` `TypedDict` gives `file_hash_projection` precise per-key types; rename the file branch's `projection` var so it isn't unified with the document branch's `dict[str, str\|None]`. |
| `api/routes/search.py`, `api/routes/knowledge.py` | Narrow `parsed.identifier` to `str` (400 on a missing identifier) before `drill_down` / returning. |
| `mcp_server/server.py` | Use a distinct name for the `str\|None` `args.get("vault")` so it doesn't clash with the `str` from `split_browse_uri`; drop the duplicate `response: dict` annotation (`no-redef`). |

No runtime behavior change for valid inputs — the new guards only affect malformed URIs (which previously would have failed later anyway).

## Verification

- **mypy**: `Success: no issues found in 117 source files` (run in a CI-equivalent clean env — mypy only, no project deps, matching `check.yml`).
- **ruff**: clean.
- `test_mcp_e2e` **76/76**; `test_graph_replace_e2e` unchanged vs `main` baseline (the lone failure is a pre-existing environmental one — no embedding/S3 locally).

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -143,13 +143,22 @@
         "is_secret": false
       }
     ],
+    "backend/CHANGELOG.md": [
+      {
+        "type": "Basic Auth Credentials",
+        "filename": "backend/CHANGELOG.md",
+        "hashed_secret": "5315196a87449ee7110cd2cce3695bdcf505e55f",
+        "is_verified": false,
+        "line_number": 372
+      }
+    ],
     "backend/mcp_server/help.py": [
       {
         "type": "Secret Keyword",
         "filename": "backend/mcp_server/help.py",
         "hashed_secret": "e5e9fa1ba31ecd1ae84f75caaa474f3a663f05f4",
         "is_verified": false,
-        "line_number": 362,
+        "line_number": 363,
         "is_secret": false
       },
       {
@@ -157,7 +166,7 @@
         "filename": "backend/mcp_server/help.py",
         "hashed_secret": "90bd1b48e958257948487b90bee080ba5ed00caa",
         "is_verified": false,
-        "line_number": 1091,
+        "line_number": 1093,
         "is_secret": false
       }
     ],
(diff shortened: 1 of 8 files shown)


kwoo24-oss:
Added a second commit fixing **backend-pytest** too: the 0.4.1 pool-deadlock fix added a `conn=` kwarg to `doc_repo.update_hash` and a required `conn=` to `DocumentService._update_locked`, but `test_document_hash_contract_unit.py`'s `_FakeDocRepo` still used the old signatures (3 failures on main). Updated the fake + the direct `_update_locked` call.

Now both CI checks pass locally:
- **check** (mypy): `Success: no issues found in 117 source files`
- **backend-pytest**: `pytest tests/ -k 'not _e2e'` → 65 passed, 58 skipped (was 3 failed, 62 passed)

PR #183, opened 2026-06-10
fix(akb): literal-aware multi-statement guard for akb_sql

### LINKED ISSUE #180: akb_sql multi-statement guard rejects literal semicolons inside string literals
## Problem

`table_service.execute_sql` enforces its single-statement boundary with a literal-blind check (`backend/app/services/table_service.py:515-516`):

```python
sql_check = rewritten.rstrip(";").strip()
if ";" in sql_check:
    return err("Multi-statement SQL is not allowed. ...", code=MULTI_STATEMENT)
```

Any semicolon inside a quoted string literal trips the guard, so a single valid statement like

```sql
INSERT INTO source_github_issues (title) VALUES ('Fix bug; refactor')
```

is rejected as multi-statement. Callers that sync real-world text (issue titles, message bodies) cannot store a semicolon losslessly. The OSS downstream collector currently works around this by replacing `;` with a fullwidth `；` before writing, which silently corrupts source-origin content — a data-fidelity boundary AKB should own, not force downstream escaping for.

The irony is that the statement has already been through `rewrite_table_names`, which is fully literal-aware: `_SQL_TOKEN_RE` plus the dollar-quote scan in `backend/app/repositories/table_data_repo.py` correctly classify string literals, quoted identifiers, and comments. The guard just doesn't use it.

## Direction

Make the multi-statement check literal-aware by reusing the existing tokenizer: a `;` only counts as a statement separator when it appears outside string literals, quoted identifiers, comments, and dollar-quoted blocks. No new parsing machinery — the rewriter already walks the exact token classes needed.

## Next Step

- Add a small `count_statement_separators(sql)` (or equivalent) next to `rewrite_table_names`, sharing `_SQL_TOKEN_RE` and `_scan_dollar_quote`.
- Replace the `";" in sql_check` test in `execute_sql` with it; keep the same `MULTI_STATEMENT` error for genuine multi-statement input.
- Regression tests: literal `;` in a single statement passes; `SELECT 1; SELECT 2` still fails; `;` inside dollar-quotes/comments/quoted identifiers passes.

## Open Questions

- None blocking. Trailing-semicolon tolerance (`rstrip(";")`) can stay as-is.

imjhyuk-oss:
Closes #180.

## What Changed

`execute_sql`'s multi-statement boundary now classifies semicolons with the same scope-awareness as the table-name rewriter: a new `count_statement_separators()` in `table_data_repo` shares `_SQL_TOKEN_RE` and `_scan_dollar_quote`, so a `;` only counts as a separator outside string literals, quoted identifiers, comments, and dollar-quoted blocks. The guard in `table_service.execute_sql` replaces the literal-blind `";" in sql_check` membership test with it; the `MULTI_STATEMENT` error and trailing-semicolon tolerance are unchanged.

## Why

A single valid statement like `INSERT ... VALUES ('Fix bug; refactor')` was rejected as multi-statement, so callers syncing real-world text could not store a semicolon losslessly — the OSS akb-collector currently substitutes a fullwidth `；` before writing, which silently corrupts source-origin content. Statement-boundary classification is AKB's contract to own; the tokenizer that already understands these token classes was one import away.

## Not Included

No change to the allowed-statement keyword check or archived-vault gating. No relaxation for genuine multi-statement input — `SELECT 1; SELECT 2` still fails with the same error.

## Validation

- `cd backend && uv run pytest tests/test_sql_multi_statement_guard_unit.py tests/test_table_identifier_unit.py tests/test_table_name_length_unit.py` — 20 passed.
- Full non-e2e backend suite: 167 passed, 61 skipped (infra-gated skips unchanged).
- New tests cover: literal `;` in a single statement passes; `SELECT 1; SELECT 2` fails; `;` inside dollar quotes, line/block comments, and quoted identifiers passes; trailing semicolons tolerated.

### DIFF
--- a/backend/app/repositories/table_data_repo.py
+++ b/backend/app/repositories/table_data_repo.py
@@ -205,6 +205,42 @@ def _scan_dollar_quote(sql: str, start: int) -> int | None:
     return end + len(tag)
 
 
+def count_statement_separators(sql: str) -> int:
+    """Count the ``;`` characters in ``sql`` that act as statement
+    separators — i.e. appear OUTSIDE string literals, quoted
+    identifiers, comments, and dollar-quoted blocks.
+
+    Shares the tokenizer (``_SQL_TOKEN_RE`` + ``_scan_dollar_quote``)
+    with ``rewrite_table_names`` so the multi-statement guard in
+    ``table_service.execute_sql`` classifies semicolons with exactly
+    the same scope-awareness as the rewriter. The previous guard was a
+    literal-blind ``";" in sql`` membership test, which rejected single
+    statements like ``VALUES ('Fix bug; refactor')`` (issue #180).
+
+    Tolerance for trailing semicolons is the caller's policy: this
+    helper reports every separator it sees.
+    """
+    count = 0
+    pos = 0
+    n = len(sql)
+    while pos < n:
+        # Same walk order as `rewrite_table_names`: manual dollar-quote
+        # scan first, then the token regex.
+        end = _scan_dollar_quote(sql, pos)
+        if end is not None:
+            pos = end
+            continue
+        m = _SQL_TOKEN_RE.match(sql, pos)
+        if not m:
+            # Should not happen — `sym` catches any character. Safety net.
+            pos += 1
+            continue
+        if m.lastgroup == "sym" and m.group() == ";":
+            count += 1
+        pos = m.end()
+    return count
+
+
 def rewrite_table_names(sql: str, table_map: dict[str, str]) -> str:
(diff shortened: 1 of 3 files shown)


skt8776-oss:
LGTM.

There is still a narrow PostgreSQL syntax edge case around escape string literals, e.g. `SELECT E'a\\'; still inside; ok'`, where the lightweight tokenizer can misclassify semicolons. I do not think that should block this PR: the fix addresses the reported normal string-literal case without adding a database/parser dependency. If we later need full PostgreSQL literal coverage, we can handle it as a follow-up by extending the tokenizer or introducing a dedicated statement-boundary helper.

Issue #215, opened 2026-06-16
Extend AKB table DDL tools with declarative unique keys, indexes, and check constraints


imjhyuk-oss:
## Background

AKB currently keeps schema changes behind dedicated DDL tools:

- `akb_create_table`
- `akb_alter_table`
- `akb_drop_table`

This boundary should remain. `akb_sql` should stay DML/query-oriented and should not become an arbitrary DDL execution surface.

However, the current DDL tools only expose basic column operations. They do not allow callers to define common database-level invariants such as uniqueness, lookup indexes, or simple check constraints. This makes clients implement important guarantees in application code, for example duplicate detection for agent event logs.

We should extend the AKB table DDL tool contract with safe declarative schema options that AKB can validate, store as metadata, and translate to the backing database.

## Proposal

Add optional schema arguments to `akb_create_table` and `akb_alter_table`:

- `unique_keys`
- `indexes`
- `check_constraints`

These should be declarative JSON objects, not raw SQL strings.

## `akb_create_table` additions

Example:

```json
{
  "parent": "akb://vaults/example",
  "name": "companion_agent_events",
  "columns": [
    { "name": "principal_id", "type": "text", "required": true },
    { "name": "workspace_id", "type": "text", "required": true },
    { "name": "session_id", "type": "text", "required": true },
    { "name": "run_id", "type": "text", "required": true },
    { "name": "seq", "type": "number", "required": true },
    { "name": "client_request_id", "type": "text" },
    { "name": "event_type", "type": "text", "required": true },
    { "name": "payload", "type": "json", "required": true }
  ],
  "unique_keys": [
    {
      "name": "companion_agent_events_run_seq_key",
      "columns": ["principal_id", "workspace_id", "session_id", "run_id", "seq"]
    }
  ],
  "indexes": [
    {
      "name": "companion_agent_events_run_order_idx",
      "columns": [
        { "name": "principal_id" },
        { "name": "workspace_id" },
        { "name": "session_id" },
        { "name": "run_id" },
        { "name": "seq", "order": "asc" }
      ]
    },
    {
      "name": "companion_agent_events_client_request_idx",
      "columns": ["principal_id", "workspace_id", "client_request_id"]
    },
    {
      "name": "companion_agent_events_event_type_idx",
      "columns": ["principal_id", "workspace_id", "event_type"]
    }
  ],
  "check_constraints": [
    {
      "name": "companion_agent_events_seq_non_negative",
      "rules": [
[...]

PR #220, opened 2026-06-17
feat(tables): declarative unique_keys + indexes on table DDL tools (#215)

imjhyuk-oss:
## Summary

Implements #215 (**PR 1 of 2**): declarative `unique_keys` + `indexes` on the table DDL tools. `check_constraints` — the raw-injection-heaviest surface — follows in a separate PR.

`akb_create_table` accepts optional `unique_keys` and `indexes` at create time; `akb_alter_table` gains `add_unique_keys` / `drop_unique_keys` / `add_indexes` / `drop_indexes`. Declarative JSON only — `akb_sql` stays DML-only.

## What's included

- **DDL primitives** (`table_data_repo`): `create_unique_constraint`, `drop_constraint`, `create_index`, `drop_index`, and a `unique_key_duplicates` preflight. Every identifier flows through `safe_ident`; index order is a closed `asc`/`desc` enum — no raw user value reaches a DDL string.
- **Deterministic name generation**: generated constraint/index names are namespaced by the physical table (index names are schema-global in PostgreSQL) and carry a `sha1` digest over the *structured* column list, so distinct column lists can never alias (`["a","b"]` vs `["a_b"]`) and names stay ≤ 63 bytes. Caller-supplied names are used verbatim.
- **Atomicity** (AC #4/#10): adding a unique key preflights existing rows for duplicates *before* `ADD CONSTRAINT`; preflight + DDL + registry write share one transaction, so a violation rolls back **both** physical schema and registry. Post-preflight races and schema-global name clashes surface as a clean `invalid_argument` (422), not a 500.
- **Stable duplicate error** (AC #2): a duplicate `INSERT` through `akb_sql` now returns a dedicated `unique_violation` code + `pg_sqlstate=23505` (was the generic `sql_error` catch-all).
- **Metadata** (AC #11): persisted on `vault_tables` (idempotent migration `037`, registered in the applier list) and exposed via `list_tables`, `akb_vault_info` introspection, and the create/alter return dicts. `init.sql` mirrors the columns for fresh installs.
- **REST parity**: `CreateTableRequest` now threads `unique_keys`/`indexes` (previously silently dropped — the REST surface could read them but not write them).

## Tests

- **DB-free unit** `test_table_constraints_unit.py` — name generation (deterministic / collision-safe / 63-byte truncation / column-list disambiguation), validation (column existence, reserved-column rejection, `asc`/`desc` enum, duplicate names), preflight query shape, registry JSON round-trip.
- **Live e2e** `test_table_constraints_e2e.sh` — AC #1–#6, #10, #11, the `akb_sql` DDL-only boundary, and the admin permission gate (AC #13). 18/18 pass against a local pgvector stack.

## Acceptance criteria

Covered: **#1, #2, #3, #4, #5, #6, #10, #11, #12, #13** (for `unique_keys` + `indexes`).
Deferred to the follow-up PR: **#7, #8, #9** (`check_constraints`) and the `akb_sql` `WITH`/CTE hardening follow-up the issue calls out.

## Note — pre-existing, out of scope

While testing the admin gate (AC #13) I noticed that a non-admin caller hitting **any** admin-gated MCP tool (`akb_alter_table`, `akb_drop_table`, `akb_grant`, …) has its `ForbiddenError` surfaced under `code=internal` by the central `call_tool` dispatch (it maps `NotFoundError`/`ValidationError`/`ConflictError`, but `ForbiddenError` falls to the generic `except Exception → internal`). The gate itself works — the call is rejected before any DDL — but the code is misleading. This predates #215 and affects many tools, so I left it for a focused follow-up (a one-line `except ForbiddenError → permission_denied` in `call_tool` would fix it uniformly).

### DIFF
--- a/.github/workflows/backend-pytest.yml
+++ b/.github/workflows/backend-pytest.yml
@@ -89,4 +89,14 @@ jobs:
         env:
           # Matches the fixture default + the dev compose override.
           AKB_TEST_DSN: postgresql://akb:akb@localhost:15432/akb  # pragma: allowlist secret
-        run: pytest tests/test_pgvector_ext_bootstrap_e2e.py -v --tb=short
+        # #215 declarative-DDL integration tests need a live PG too (index
+        # ASC/DESC physical order, migration 037 upgrade path, add-unique-key
+        # concurrency race). They skip in the DB-free unit job above, so gate
+        # them here where AKB_TEST_DSN is reachable.
+        run: >
+          pytest
+          tests/test_pgvector_ext_bootstrap_e2e.py
+          tests/test_migration_037_upgrade_unit.py
+          tests/test_index_order_unit.py
+          tests/concurrency/test_alter_unique_race_unit.py
+          -v --tb=short
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -157,6 +157,7 @@ jobs:
             test_pg_rbac_e2e.sh
             test_graph_replace_e2e.sh
             test_jwt_revocation_e2e.sh
+            test_table_constraints_e2e.sh
           )
           TOTAL_PASS=0
           TOTAL_FAIL=0
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -166,7 +166,7 @@
         "filename": "backend/mcp_server/help.py",
         "hashed_secret": "90bd1b48e958257948487b90bee080ba5ed00caa",
         "is_verified": false,
-        "line_number": 1096,
+        "line_number": 1111,
         "is_secret": false
       }
     ],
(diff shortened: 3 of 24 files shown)


imjhyuk-oss:
Filed the pre-existing `ForbiddenError` → `code=internal` finding (surfaced by the AC#13 permission test) as a tracked issue: #221. Kept out of this PR per scope.

imjhyuk-oss:
Pushed review fixes (e907a61) from a deep adversarial re-review: NULLS-DISTINCT-aware duplicate preflight (no longer blocks a valid UK on a nullable column), duplicate-column / mixed-case column refs now clean `invalid_argument` (were DDL 500s), and column rename/drop now keeps the unique_key/index registry metadata consistent. e2e hardened 18→28 cases. Release prep per CLAUDE.md: 0.8.15→**0.8.16** + CHANGELOG, and registered `test_table_constraints_e2e.sh` in the e2e.yml shell-e2e gate (it never ran in CI before). Note: #220 and #222 both touch `pyproject.toml`/`CHANGELOG.md`/`e2e.yml` — whichever merges second needs a trivial rebase (versions assume #220→0.8.16 lands before #222→0.8.17).

PR #222, opened 2026-06-17
fix(mcp): map access-guard errors to stable codes instead of code=internal (#221)

### LINKED ISSUE #221: bug(mcp): access-gated tools return ForbiddenError as code=internal instead of permission_denied
## Summary

A non-admin (or otherwise unauthorized) caller hitting any **access-gated MCP tool** is correctly *rejected*, but the failure is surfaced under the generic **`code=internal`** envelope instead of a stable permission code (e.g. `permission_denied` / HTTP 403). Clients can't reliably distinguish "you lack the role" (a 4xx the caller should handle) from "the server crashed" (a 5xx).

## Impact

- Misleading error taxonomy for every admin/writer-gated MCP tool: `akb_alter_table`, `akb_drop_table`, `akb_grant`, `akb_revoke`, `akb_transfer_ownership`, `akb_archive_vault`, `akb_delete_vault`, etc.
- The access **gate itself works** — the call is rejected *before* any side effect — so this is a contract/observability bug, not a security hole.
- Surfaced while adding permission coverage for #215 (see PR #220, AC #13): a reader calling `akb_alter_table` is rejected, but with `code=internal`.

## Root cause

`check_vault_access(...)` raises `ForbiddenError` (a subclass of `AKBError`) — e.g. `backend/app/services/access_service.py:121` (`Requires '<role>' role on vault ...`). The handlers call it **outside** their `try` blocks and only catch `NotFoundError` / `ValueError` / `ValidationError` / `ConflictError`. The central dispatch in `call_tool` (`backend/mcp_server/server.py:1258`) has no `except ForbiddenError`, so it falls through to the last-resort handler:

```python
# backend/mcp_server/server.py ~1266
    except Exception as e:
        ...
        envelope = err(str(e), code=INTERNAL)   # ~1273  ← ForbiddenError lands here
```

So `ForbiddenError` → `code=internal`. (`PERMISSION_DENIED = "permission_denied"` already exists in `app/util/errors.py` and is used by the `akb_sql` PG-ACL path, so the taxonomy slot exists — it's just not wired for `ForbiddenError`.)

## Reproduction

1. User A creates a vault and a table.
2. User A grants user B `reader`.
3. User B calls `akb_alter_table` (or `akb_drop_table`, `akb_grant`, …) on that vault.
4. Expected: a stable `permission_denied` (403) envelope. Actual: `{"error": "Requires 'admin' role on vault ...", "code": "internal"}`.

## Suggested fix

Map `ForbiddenError` to the existing permission code at the single central dispatch point, so every gated tool benefits uniformly:

```python
# in call_tool (or _dispatch), before the generic `except Exception`
    except ForbiddenError as e:
        envelope = err(str(e), code=PERMISSION_DENIED)
        audit_log.record_tool(name, arguments, user, envelope)
        return [TextContent(type="text", text=json.dumps(envelope, ensure_ascii=False, default=str))]
```

(Requires importing `ForbiddenError` + `PERMISSION_DENIED` in `server.py`.)

**Compatibility note:** existing permission e2e tests (`test_edit_e2e.sh`, `test_mcp_isolation.sh`, `test_security_edge_e2e.sh`) assert on the message *text* (`forbidden` / `denied` / `role`), not the code, so they keep passing after the fix. A small e2e asserting `code == "permission_denied"` should be added alongside.

## Scope note

Out of scope for #215 / PR #220 (which is focused on declarative `unique_keys` + `indexes`); filed separately so the cross-cutting dispatch change is reviewed on its own.

imjhyuk-oss:
## Summary

Fixes #221. Access guards (`check_vault_access`) raise `ForbiddenError` / `NotFoundError` **outside** the per-handler try/except blocks, so the central `call_tool` dispatch's last-resort `except Exception → code=internal` mislabeled every permission denial / missing-vault reference on admin/access-gated tools as a 500-shaped `internal` instead of the 4xx it is. The gate always worked (the call is rejected before any side effect) — only the surfaced **code** was wrong, so clients couldn't tell "you lack the role" from "the server crashed".

## Fix

Extract `_exception_envelope(e)` (pure, unit-testable) and use it in `call_tool`:

| exception | code (was `internal`) |
|---|---|
| `ForbiddenError` | `permission_denied` |
| `NotFoundError` | `not_found` |
| `ConflictError` | `conflict` |
| `ValidationError` | `invalid_argument` |
| anything else | `internal` |

One central point → fixes it uniformly for every gated tool (`akb_alter_table`, `akb_drop_table`, `akb_grant`, `akb_vault_info`, …). An unmapped `AKBError` subclass still falls to `internal` (no silent reclassification of an unknown error type).

## Tests

- `test_mcp_error_envelope_unit.py` — the mapping, plus unmapped `AKBError` / unknown exception staying `internal`, and the canonical `{error, code}` envelope shape.
- `test_forbidden_permission_code_e2e.sh` — reader → `alter_table` / `drop_table` / `grant` all return `permission_denied`; a missing vault → `vault_info` returns `not_found`; admin alter still succeeds (no regression).

Verified locally (Python 3.14 venv): **7 unit + 7 e2e pass**; `ruff` / `mypy --python-version 3.14 app/ mcp_server/` / `bandit --severity-level medium` / `detect-secrets` all clean.

## Note

No version bump / `CHANGELOG.md` entry included here — this repo batches those into dedicated `chore(backend): bump to X.Y.Z + CHANGELOG` commits, and a per-PR bump tends to conflict on the fast-moving `main`. Happy to add one if you'd prefer it in-PR.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -158,6 +158,7 @@ jobs:
             test_graph_replace_e2e.sh
             test_jwt_revocation_e2e.sh
             test_table_constraints_e2e.sh
+            test_forbidden_permission_code_e2e.sh
           )
           TOTAL_PASS=0
           TOTAL_FAIL=0
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "backend/CHANGELOG.md",
         "hashed_secret": "5315196a87449ee7110cd2cce3695bdcf505e55f",
         "is_verified": false,
-        "line_number": 2782
+        "line_number": 2799
       }
     ],
     "backend/mcp_server/help.py": [
--- a/backend/CHANGELOG.md
+++ b/backend/CHANGELOG.md
@@ -5,6 +5,23 @@ the `akb-mcp` stdio proxy. This changelog tracks the backend
 specifically; the proxy has its own log in
 `packages/akb-mcp-client/CHANGELOG.md` and a separate version stream.
 
+## 0.8.17 — 2026-06-17  *(fix — MCP access-guard errors get stable codes, not code=internal, #221)*
+
+`check_vault_access` raises `ForbiddenError` / `NotFoundError` *outside* the
+per-handler try/except blocks, so the central `call_tool` dispatch's last-resort
+`except Exception -> code=internal` mislabeled every permission denial /
+missing-vault reference on admin/access-gated tools (`akb_alter_table`,
+`akb_drop_table`, `akb_grant`, `akb_vault_info`, …) as a 500-shaped `internal`
+instead of the 4xx it is. The gate always worked (the call is rejected before any
+side effect) — only the surfaced code was wrong.
+
+New pure helper `exception_envelope(e)` in `app/util/errors.py` maps known
+`AKBError` subclasses to their canonical codes (`ForbiddenError ->
+permission_denied`, `NotFoundError -> not_found`, `ConflictError -> conflict`,
+`ValidationError -> invalid_argument`); anything else stays `internal`. Living in
+the error-shape module (no import-time side effects) keeps it unit-testable
+without importing the MCP server. New `test_mcp_error_envelope_unit.py` +
+`test_forbidden_permission_code_e2e.sh` (registered in the CI shell-e2e gate).
 ## 0.8.16 — 2026-06-17  *(feat — declarative unique keys + indexes on the table DDL tools, #215)*
 
 `akb_create_table` accepts optional declarative `unique_keys` and `indexes`;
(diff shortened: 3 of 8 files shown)


imjhyuk-oss:
Force-pushed 3f5f258. Fixes the earlier `backend unit tests` red: the unit test imported `mcp_server.server`, whose module import touches `git_storage_path` (`/data`) and failed at collection in CI. Moved the mapper to the side-effect-free `app/util/errors.py` as `exception_envelope` so it is unit-testable without importing the server. Added release prep: 0.8.15→**0.8.17** + CHANGELOG + registered `test_forbidden_permission_code_e2e.sh` in the e2e.yml gate. See the #220 note re: shared-file rebase on merge order.

PR #233, opened 2026-06-22
feat(frontend/graph): graph viewer Phase 0+1 — live sim, degree circles, typed edges, local-first nav + viewport-culling/LOD progressive rendering

kwoo24-oss:
Research-driven rebuild of the knowledge-graph viewer's render + interaction model so it reads like a daily tool (Obsidian-grade), not a stuck demo. Reuses the existing data/state layer + `react-force-graph-2d` per the accepted design doc (`docs/designs/graph-viewer-v2/00-overview.md`). Driven by a multi-agent reference pass (Obsidian/Roam/Logseq/Foam/Neo4j Bloom/Gephi) and hardened by a 15-agent adversarial review (findings verified + fixed).

## Phase 0 — correctness ("renders right")
- **Live damped simulation.** Removed pin-on-settle (which fixed `fx/fy` on *every* node at first settle, freezing the layout so overlaps never relaxed and expand/cluster-toggle had to un-pin and reheat into the same frozen mess). Added `forceCenterPull` — a weak spring to the origin giving the layout a real equilibrium so charge can't ratchet clusters outward. Only **user drags** pin nodes now (the Obsidian/d3 model; verified against force-graph 1.51.2 that clicks neither pin nor reheat).
- **Doc → graph link fixed.** The document page emitted `?focus=<akb-uri>` but the graph only read `?entry=<doc-id>`, so "Open in graph" never focused — it dumped the whole vault. Now emits `?entry=<path>` + a back-compat `focus` alias so already-shipped links land focused.
- Reduced-motion Resume reheat gated on the live `frozen` flag; expand seeds new neighbors in a deterministic spiral around the parent.

## Phase 1a — render (topology as the primary channel)
- Square + `D/T/F` glyph nodes → **degree-sized circles** (`sqrt` sizing); kind carried by **shape** (document = circle, table = rounded square, file = dashed-ring) instead of a glyph illegible at overview zoom.
- **Selection is the single teal signal** (border + halo + incident edges + arrowheads + label); orange survives only as the user-pin marker, so the two never compete.
- **Typed edges** read at a glance: a structural (solid/darker/thicker) vs associative (dashed/muted/thinner) ramp + per-relation dash; arrowheads inherit the edge encoding. Hubs paint first so their labels win the collision cull. Cluster hues capped to cat-1..4.

## Phase 1b — navigation / a daily driver
- **Depth (hops)** shown in both modes + persisted; **Orphans toggle** (+ "N unconnected") tightens zoom-to-fit onto the connected core; **Hubs list** (degree-ranked, click → select + center) — a visible way into a large graph and a keyboard path.
- Relations are now a **swatch list in each relation's own encoding** so the rail doubles as the legend; humanized labels; `aria-live` announces the selection + its degree.

## Review fixes (HIGH)
- Pin physics were decoupled from the `pinned` Set (only a drag wrote `fx/fy`); added a `pinned`-keyed effect that reconciles physics so menu/panel Pin actually freezes and Unpin releases.
- Orphan toggle / counts / aria-degree read raw edge endpoints that force-graph mutates string→object in place, which blanked the canvas on toggle — normalized every endpoint read with `endpointUri`.

## Verification
`design:check` / `typecheck` / `lint` (0 errors) / **310 tests** (+4) green; production build OK; local docker frontend rebuilt + smoke-checked (0 console errors).

## Not in scope (Phase 2, backend)
Degree-ranked `/graph/overview` with honest "N of M", `/graph/health` (orphans/hubs), single-call backend BFS, implicit-`links_to` promotion. Tracked for a follow-up.


kwoo24-oss:
Added a 5th commit: **viewport culling + LOD progressive rendering** (`f3e8990`).

The viewer now renders proportional to the visible screen ratio instead of painting all N nodes every frame:
- **Overview** (whole graph) → only the strongest ~30 hubs draw (a readable constellation) + their skeleton edges.
- **Zoom into a region** → leaves / labels / edges progressively appear (a COUNT-target degree floor, hysteresis-stable).
- **Pan** → off-screen nodes + edges are culled (`nodeVisibility`/`linkVisibility` skip paint *and* hit-test; the d3 sim keeps the full set, so layout is undisturbed).
- Gated OFF below 300 nodes (small graphs unchanged).

New pure `lod.ts` (15 unit tests). Verified hands-on via Playwright on a 400-node synthetic vault: overview→hubs, zoom→leaves, 8-node graph unchanged, console clean.

kwoo24-oss:
Large follow-up iteration (commits f922116 → e9b95e6), all hands-on-verified on a 344-node test vault via Playwright + gated (design:check / typecheck / lint 0 / 330 tests):

**LOD, refined from feedback**
- Leaves are no longer *hidden* at overview (read as "where did my nodes go?") — below-top-N nodes render as small dim cluster-tinted dots (the structure stays honest; hubs dominate). Fixed the budget too: an exact top-N set, not a degree floor (a floor over-admitted on ties).
- Fit anchor derived from the live graph bbox (self-calibrates after expand/cluster/filter; no stale anchor, no settle-timing race), cached so the per-zoom cull is O(1).
- Depth (hops) disabled + dimmed on the whole-graph view (it's a no-op without a focus).

**xhigh /code-review (64 agents, verified) — confirmed bugs fixed**
- LOD band went stale after a non-camera relayout (cluster-toggle/expand) → recompute on every settle.
- Edges crossing the viewport with both endpoints off-screen were culled → segment-vs-rect test.
- Focus/pinned neighborhoods could show a missing edge → force-show edges whose both endpoints are forced.
- Hit-test area now matches the degree-driven radius (tiny dots stay clickable).
- Dedup: shared `degreeMap` / `connected` set / `KindSwatch`+`RelationSwatch`.

**ui-ux-pro-max plan (A+B+C)**
- Polish: native-input teal tints + focus rings, context-menu ring, swatch sharing.
- Completeness: empty-graph "Reset filters", "Unhide all (N)", recallable gesture hint, detail-panel "Focus", canvas-filling loading.
- **Impact-analysis mode**: toggle → directed dependency (teal) + dependent (orange) cones over structural edges, rest dimmed — "what depends on this before I edit it".

PR #262, opened 2026-07-06
[codex] Add Storybook page scenario coverage

younglokim-oss:
## What changed

- Added Storybook 10 for the frontend with React Vite, Docs, A11y, Vitest browser tests, MSW, and a light/dark toolbar.
- Added route-backed AKB page stories that mirror the app route tree, including auth, app home/settings, search, vault resources, vault admin, documents/publications, and recovery states.
- Added shared Storybook MSW fixtures and route-shell assertions so page stories verify when app chrome, vault rail, and public layouts should or should not appear.
- Updated the auth callback page to read query parameters from React Router location, which keeps SSO callback stories working under MemoryRouter while preserving browser behavior.

## Why

The frontend did not have Storybook coverage. This gives AKB a usable Storybook setup plus scenario-level page stories for visual review, accessibility checks, and interaction testing.

## Impact

Developers can run Storybook locally, build it for review, and execute Storybook Vitest coverage separately from unit tests. The app runtime behavior is unchanged except for the router-compatible auth callback query read.

## Validation

- `cd frontend && pnpm run typecheck`
- `cd frontend && pnpm run test:storybook`
- `cd frontend && pnpm run build-storybook`
- `cd frontend && pnpm run design:check && pnpm run typecheck && pnpm run lint && pnpm run test && pnpm run build`
- `$autoreview` clean: no accepted/actionable findings reported


younglokim-oss:
Reviewer note: this PR is primarily Storybook setup + scenario coverage, but a few adjacent existing files changed to make that setup behave correctly without changing the app/CI contract unexpectedly.

- `frontend/src/pages/auth-callback.tsx`
  - Changed the callback page to read query params from React Router's `useLocation().search` instead of `window.location.search`.
  - Reason: Storybook renders page scenarios under `MemoryRouter`; `/auth/callback?code=...` exists in router state, while `window.location.search` is Storybook's own `?path=/story/...` URL. This keeps production `BrowserRouter` behavior equivalent while making the route-backed SSO callback stories accurate.

- `frontend/src/index.css` and `frontend/DESIGN_SYSTEM.md`
  - Only comments/docs changed, not runtime tokens or styles.
  - Reason: removed placeholder-looking Tailwind arbitrary class examples such as `z-[var(--z-*)]` / `rounded-[var(--radius-*)]` from prose. They were documentation shorthand, not real class names, so the wording now describes the token usage without looking like extractable classes.

- `frontend/pnpm-workspace.yaml`
  - Added `allowBuilds: esbuild: true` for the frontend package install policy.
  - Reason: Storybook/Vite/Vitest dependencies require the esbuild build script to be explicitly approved under pnpm's current build-script policy. The frontend is installed as its own pnpm project in CI, so the setting is scoped under `frontend/`.

- `scripts/check.sh`
  - Changed the frontend test command from direct `vitest run` to `pnpm run test`.
  - Reason: after this PR, Vitest has separate `unit` and `storybook` projects. Direct `vitest run` executes both, which would run the Storybook browser project in CI and require Playwright browser installation. `pnpm run test` maps to `vitest run --project unit`, so CI keeps running only the existing unit/RTL/MSW gate. Storybook tests remain available separately via `pnpm run test:storybook` and are intentionally not added to CI in this PR.

Validation after these adjustments:
- `pnpm run test`
- `pnpm run test:storybook`
- `pnpm run build`
- `pnpm run build-storybook`
- PR CI passed: static analysis, backend unit tests, pgvector e2e, shell e2e

kwoo24-oss:
## Review — Storybook page scenario coverage

**No blocking issues.** Additive Storybook setup; no backend/proxy boundary touched (0 changes under `backend/` or `akb-mcp-client`). Approving with a few follow-ups.

### Only runtime change — verified equivalent in prod
`frontend/src/pages/auth-callback.tsx`: `window.location.search` → `useLocation().search`.
- App renders under `BrowserRouter` (`main.tsx:69`), so `useLocation().search === window.location.search` in production. The SSO callback is always a full-page redirect from the IdP, so the router location is already populated on mount. Behaviorally equivalent. ✅
- The `if (ran.current) return;` guard (`auth-callback.tsx:23`) is intact, so the newly-added `routerLocation.search` effect dependency cannot trigger a second code redeem. ✅
- `rawRedirect` still resolves against `window.location.origin`, which is correct (origin is router-independent).

### CI contract
- `scripts/check.sh`: `npx --no-install vitest run` → `pnpm run test` (= `vitest run --project unit`). Keeps the Playwright-backed Storybook project out of the CI gate — reasonable. **Note:** the new Storybook interaction/a11y tests are therefore *not* gated in CI (only `pnpm run test:storybook`, run manually). The scenario coverage this PR adds won't catch regressions automatically. If that's intended, fine — worth wiring into a dedicated workflow later.
- `vitest.config.ts`: split into `unit` + `storybook` projects; unit config preserved + `storybook-static/**` excluded. LGTM.

### Build policy — one question
`package.json` drops `pnpm.onlyBuiltDependencies: ["msw"]`; new `frontend/pnpm-workspace.yaml` adds `allowBuilds: { esbuild: true }`. esbuild approval is needed for the vite/vitest toolchain. **Q:** was removing `msw` from build-approved deps intentional? Evidence says it's safe (msw needs no install script; the worker is committed and the unit tests that use msw pass) — just confirming nothing relied on it.

### Maintenance (non-blocking)
- `page-route-shell.tsx` `AkbRouteTree` hand-mirrors the real route tree in `main.tsx`. It's an exact match today, but nothing enforces parity — routes added to `main.tsx` will silently be missing from the stories. Consider a comment pointing at `main.tsx` as the source of truth, or sharing a single route array.
- `preview.tsx` uses `onUnhandledRequest: "bypass"` — a missing MSW handler silently hits the real network instead of failing loudly. `"warn"` would surface such gaps.

### Security — clean
No secrets (matches are UI labels / mock 401·403 bodies). All fixture URLs are `*.example.test` / `example.com` / `localhost`. The detect-secrets exclusion of the committed `mockServiceWorker.js` is justified (integrity checksum). `setToken("mock")` runs only in the Storybook decorator.

PR #278, opened 2026-07-14
feat(auth): add delegated file upload capability

imjhyuk-oss:
## Summary
- add action-limited managed Vault write grants while preserving existing wildcard grants
- require an upload-only service key plus an active human AKB session for managed file upload and confirmation
- preserve both actor identities in bounded file audit fields
- add migration and regression coverage for compatibility, denial, and token-profile invariants

## Security boundaries
- `file_upload` grants require a non-admin service account, service key, exact `write` scope, and exactly one Vault
- delegated identities must be active human accounts with explicit writer-or-higher Vault membership
- limited grants do not authorize document, table, delete, SQL, or lifecycle writes

## Verification
- focused policy/delegation suite: 71 passed
- full backend suite: 830 passed, 24 skipped
- Ruff, mypy, and `uv build`: passed

### DIFF
--- a/backend/app/api/deps.py
+++ b/backend/app/api/deps.py
@@ -1,5 +1,6 @@
 """Shared dependencies for API routes."""
 
+from dataclasses import dataclass
 from typing import NoReturn
 
 from fastapi import HTTPException, Request, Security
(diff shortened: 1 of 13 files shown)


imjhyuk-oss:
Blocking contract finding before merge: the initial transition from an unmarked Vault is not atomic. `set_vault_write_policy()` commits the policy first, while `add_vault_write_grant()` requires an already-marked Vault and commits separately. Therefore any Vault with existing writers enters a fail-closed outage window between marking and restoring wildcard grants; pre-seeding is impossible because the FK/grant API requires the policy.

Please add a system-admin bootstrap operation that validates all supplied token/action grants and commits the policy plus the complete initial grant set in one DB transaction. Keep the existing single-policy/single-grant APIs for later mutations. Add a PostgreSQL/route regression test proving that a validation or insertion failure rolls back both policy and grants, and that no committed policy-without-grants intermediate state is observable.

imjhyuk-oss:
Re-review complete: the atomic bootstrap closes the initial fail-closed cutover window. The policy, complete initial grant set, and one audit event share the Vault-locked PostgreSQL transaction; injected second-grant failure proves full rollback. Focused 68/75 lanes and all four GitHub checks are green. No unresolved High/Medium finding.

PR #290, opened 2026-07-20
fix(sql): cap akb_sql result rows to bound event-loop stall

kwoo24-oss:
## Problem

`akb_sql` / REST `/tables/{vault}/sql` returned the **entire** SELECT result: `await conn.fetch(sql)` then `[_coerce_row(r) for r in rows]` + JSON serialisation — all **pure-Python CPU on the single event loop with no `await`**. A large SELECT (e.g. `SELECT * FROM generate_series(1, 2e6)`) stalls `/livez` past the liveness probe → **503**, from a **single request** (no concurrency needed).

Found by the 2026-07-20 event-loop-blocking audit. Measured stall scales linearly:

| rows | /livez stall | resp size |
|---|---|---|
| 250k | 0.88s | 55MB |
| 500k | 1.49s | 111MB |
| 1M | **3.07s** | 222MB |

→ ~1.7M rows crosses the 5s probe.

## Fix

Cap the **returned** rows at `settings.akb_sql_max_rows` (default 10000) and set `truncated: true` so callers paginate. The statement still executes identically — **data-modifying CTEs (`WITH ... UPDATE ... RETURNING`) are unaffected** — only the returned/serialised row count (what stalls the loop) is bounded.

## Verification

Local docker compose (single uvicorn worker = prod `replicas=1`), 1M-row SELECT:

| | before | after |
|---|---|---|
| /livez max stall | **3.07s** | **0.055s** |
| response | 222MB | 2.2MB |
| envelope | — | `total=10000, truncated=true` |
| 5-row control | — | `total=5, truncated=false` |

Adds `test_akb_sql_row_cap_unit.py` — DB-free (fake pool): over-cap → truncated, under/at cap → not truncated.

## Scope note (follow-up)

This bounds the **event-loop stall** (the 503) by capping coerce + serialize. **Peak fetch memory is still the full result set** — streaming with an asyncpg cursor to bound memory too is a follow-up, deferred because a cursor's interaction with data-modifying CTEs needs its own validation. Sibling companion PR #289 offloads the git-read blocking paths from the same audit.

### DIFF
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -91,6 +91,14 @@ class Settings(BaseModel):
     pg_pool_min_size: int = 2
     pg_pool_max_size: int = 30
 
+    # Max rows a single `akb_sql` / REST SQL SELECT may return. The result is
+    # coerced + JSON-serialised on the single event loop, so an unbounded
+    # SELECT (e.g. `SELECT * FROM generate_series(1, 2e6)`) stalls /livez past
+    # the liveness probe → 503. Over-limit results are truncated and flagged
+    # (`truncated: true`) so callers paginate. Statement execution is
+    # unchanged — only the returned row count is capped.
+    akb_sql_max_rows: int = 10000
+
     # Git storage root (bare repos live here)
     git_storage_path: str = "/data/vaults"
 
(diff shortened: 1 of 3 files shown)


kwoo24-oss:
일단 원래대로 되돌립니다(revert per request) — 접근 재검토 예정. 브랜치/커밋은 보존, 필요 시 reopen.

PR #293, opened 2026-07-20
fix(sql): keep akb_sql off the event-loop critical path

kwoo24-oss:
## Problem

AKB runs a **single uvicorn worker on a single event loop** (`replicas=1`, no `--workers`). Any request handler that does multi-second synchronous work with no `await` starves the loop, so `/livez` can't be scheduled → the kubelet liveness probe times out → the pod is SIGKILLed → **total outage** (not just a slow request).

`akb_sql` is arbitrary SQL, so a caller can trivially return a huge result set (`SELECT * FROM generate_series(1, 1e6)`). Two spots on that path are pure-Python CPU with no yield:

1. **Row coercion** — `[_coerce_row(r) for r in rows]` in `user_sql_executor.py`.
2. **JSON serialisation** — FastAPI's single `json.dumps` of the response envelope.

Measured locally against the docker-compose harness (single worker = prod): a 1M-row SELECT stalled `/livez` for **~3.07 s** → probe-timeout territory.

## Fix

Interleave `await asyncio.sleep(0)` through both hot spots so the loop keeps servicing health probes and other requests **without truncating the result** — `akb_sql` is a free-SQL tool; callers bound their own rows with `LIMIT`, and the client still receives one ordinary JSON document.

- `_coerce_rows_yielding()` coerces rows in `settings.akb_sql_coerce_batch` (2000) chunks, yielding between batches.
- `_stream_json()` serialises the envelope with `json.JSONEncoder.iterencode`, flushing ~64 KiB chunks via `StreamingResponse`, yielding every 1000 pieces.

After: same 1M-row SELECT → `/livez` max **~0.146 s**, response `total=1000000 items=1000000`, valid JSON, no `truncated` key.

## API contract

**Unchanged.** The OpenAPI schema (`POST /tables/{vault}/sql` → 200 `AkbSqlEnvelope`, `application/json`) is generated from route metadata, not the runtime return value — `test_rest_openapi_contract_unit.py` still passes. The body is byte-for-value identical (same keys, `ensure_ascii=False`, compact separators — matching Starlette's default `JSONResponse`). Only the transport differs: chunked (no `Content-Length`), invisible to any client that reads the body / calls `.json()`.

## Tests

- New `test_akb_sql_nonblocking_unit.py` (DB-free): coercion yields + preserves all rows in order; small-result fast path; `_stream_json` emits one valid document == `json.dumps`, in multiple chunks.
- Updated `test_tables_route_unit.py` to assert the route streams the same envelope.
- Full `pytest -k 'not _e2e'`: **701 passed, 167 skipped, 0 failed** (against live PG/MinIO).

Part of the single-process event-loop hardening series (companion PRs: bcrypt offload #283, git-read offload #289, S3 timeouts).

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


kwoo24-oss:
Squash-merged to `main` in 3c57731 (local integrate + full validation before push): non-e2e unit **712/0**, MCP e2e **88/0**, security-edge e2e **66/0**, plus per-PR reproduction tests. Closing; branch retained. 🤖 via local integrate flow

PR #294, opened 2026-07-20
fix(s3): bound every S3 call with connect/read timeouts + short retry

kwoo24-oss:
## Problem

boto3/botocore default to a **60 s connect AND 60 s read timeout with NO retries**. A stalled MinIO/S3 (network blip, cold bucket, dead/mis-configured endpoint) then blocks the caller for up to 60 s.

Several S3 primitives run **on the single event loop** (public raw-file read, publication-snapshot put/read, upload HEAD-confirm, bucket cold-start `ensure_bucket`). On our single-worker/single-loop backend (`replicas=1`), one such stall starves `/livez` → liveness probe timeout → pod SIGKILL → **total outage**.

## Fix

`make_client()` now stamps `settings.s3_*` onto the boto config for **every** S3 client (primary + presign + audit store):

- `connect_timeout` ≤ 3 s
- `read_timeout` ≤ 10 s
- `retries={"max_attempts": 2, "mode": "standard"}` — a short retry for transient blips.

botocore's `max_attempts` is the **retry** count, so the default 2 = 3 total attempts → worst-case read hang **3 × 10 = 30 s** (down from an unbounded 60 s, and now recoverable); a connect failure is bounded to 3 × 3 = 9 s.

## Verification

Against a blackhole endpoint (`192.0.2.1`, TEST-NET-1 — SYN dropped), a `head_bucket` now raises `ConnectTimeoutError` in **~9.7 s**, not 60 s.

## Tests

- New `test_s3_boto_timeouts_unit.py` (network-free): `make_client` applies the settings timeouts + retry mode; guards the ceiling (`total_attempts × read_timeout ≤ 60`); proves the values come from `settings` (not a hardcode) via monkeypatch.
- Full `pytest -k 'not _e2e'`: **701 passed, 167 skipped, 0 failed** (against live PG/MinIO).

Complementary to offloading the specific on-loop S3 calls to `asyncio.to_thread` (follow-up); this PR bounds the *duration* of any S3 stall globally in one place. Part of the single-process event-loop hardening series.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

PR #295, opened 2026-07-20
fix(k8s): widen liveness tolerance so a transient loop stall can't SIGKILL the sole replica

kwoo24-oss:
## Problem

The backend runs a **single event loop at `replicas=1`**, so a liveness SIGKILL is a **total outage**. But a healthy loop can be briefly unschedulable during a big `akb_sql`, an indexing burst, or a git-subprocess fan-out. The old liveness (`timeout 5s`, `failureThreshold 3`, `period 30s` → ~90s) restarted the only pod on such **transient** stalls — the exact mechanism behind the 2026-07-20 outage (loop stall → `/livez` `context deadline exceeded` → kubelet SIGKILL, exit 137).

## Change (`deploy/k8s/backend.yaml`)

| Probe | Before | After | Effect |
|---|---|---|---|
| liveness | timeout 5, failureThreshold 3, period 30 | timeout **8**, failureThreshold **6**, period 30 | ~**180s** sustained unresponsiveness before assuming a real deadlock |
| readiness | timeout 5, failureThreshold 3, period 15 | timeout **8**, failureThreshold 3, period 15 | stays honest — still sheds a stalled pod in ~45s, just doesn't flap on one slow `/readyz` |
| startup | (unchanged) | (unchanged) | 300s startup budget |

**Readiness intentionally stays tight** (traffic gate = honest), **liveness widens** (restart = last resort). This is the documented K8s/SRE pattern for single-worker async apps: liveness detects a *dead* loop, not a *busy* one.

## Why in the manifest, not `kubectl patch`

`deploy-internal.sh` re-applies this manifest and clobbers ad-hoc runtime patches (that's why the #244-era probe tuning never stuck). The internal overlay (`deploy/k8s/internal/`) does **not** override probes, so this base value is authoritative and survives the next deploy.

## Relationship to the other hardening PRs

Probe tolerance stops **pointless restarts on isolated transient stalls**; it does NOT fix sustained/concurrent blocking. The real cure is *not blocking the loop* — shipped separately: akb_sql streaming (#293), S3 timeouts (#294), bcrypt offload (#283, merged), git-read offload (#289). This PR is the safety-net layer.

No code paths changed; manifest-only. YAML validated (5 docs parse; liveness tolerance computes to 180s).

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/deploy/k8s/backend.yaml
+++ b/deploy/k8s/backend.yaml
@@ -126,20 +126,31 @@ spec:
             periodSeconds: 5
             timeoutSeconds: 3
             failureThreshold: 60
+          # Readiness gates traffic and MUST stay honest — shed a stalled pod
+          # from Service endpoints quickly. Keep failureThreshold small; only
+          # widen timeout so a single slow (not dead) /readyz doesn't flap the
+          # endpoint. ~3×15s = 45s of sustained unreadiness before removal.
           readinessProbe:
             httpGet:
               path: /readyz
               port: 8000
             periodSeconds: 15
-            timeoutSeconds: 5
+            timeoutSeconds: 8
             failureThreshold: 3
+          # Liveness RESTARTS the pod (SIGKILL) — with replicas=1 that is a
+          # total outage, so it must fire ONLY on a genuinely dead loop, never
+          # on a transient multi-second stall (a big akb_sql, an indexing
+          # burst). The single event loop can be briefly unschedulable while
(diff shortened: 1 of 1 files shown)


kwoo24-oss:
Squash-merged to `main` in 9843e27 (local integrate + full validation before push): non-e2e unit **712/0**, MCP e2e **88/0**, security-edge e2e **66/0**, plus per-PR reproduction tests. Closing; branch retained. 🤖 via local integrate flow

PR #313, opened 2026-07-27
fix(mcp): akb_grep(replace=) required only read scope — promote to write

kwoo24-oss:
## What

`akb_grep` was mapped to `akb:vault:read` in `_TOOL_SCOPES`, but it accepts an
optional argument that rewrites every matching document. That mapping is a real
enforcement gate — it is consulted for both OAuth access tokens and PAT scopes —
so a token deliberately scoped down to `read` passed the gate and reached the
mutating path.

## Severity

**Not privilege escalation.** The handler still requires the vault `writer`
role, so a vault *reader* cannot write.

What it defeated is the reason read-only tokens exist: handing an agent a read
PAT is supposed to mean it cannot change anything, and for a caller who *is* a
vault writer, it did not.

**Release impact — two different answers.** The scope bypass affects
unreleased `main` only: the gate it defeats was introduced after
`backend-v0.9.4`, so no tagged release carries it, and operators building from
`main` get the fix with the same `git pull`.

The audit-classification defect fixed alongside it **does** predate
`backend-v0.9.4`. There, `akb_grep` and `akb_publication_snapshot` sit in
`audit_log._READ_ONLY_TOOLS`, which is skipped entirely when `log_reads` is
false — while the shipped `app.yaml.example` states that writes are always
logged.

Auditing is off by default (`audit.enabled: false`) and `log_reads` defaults
to true, so this reaches only deployments that turned auditing **on** and read
logging **off**. **Those operators should assume mass replacements and
publication snapshots are absent from their audit stream on 0.9.4 and
earlier.** Everyone else is unaffected.

## Fix

`_required_scope(name, args)` promotes a read-grade tool to write when the call
carries a declared mutating argument; both gates use it.

The rule lives in `_ARG_WRITE_TRIGGERS` rather than a branch inside `_dispatch`,
so the next tool that grows such an argument has one obvious place to declare
it. The trigger tests for presence rather than truthiness, so a falsy-but-
supplied value is still classified as a write.

**Rejected:** mapping the whole tool write-grade. That would take literal and
regex search away from every read-only agent, and `akb_search` does not cover
it. A separate read/write tool split is the durable answer but changes the
published MCP `inputSchema`, so it belongs in the contract-convergence pass —
the same call `akb_sql`'s comment already records.

## Why CI was green

The existing guard asserted that every tool *has* a scope mapping, never that
the mapping is *correct*. `akb_grep` had one, so it passed.

Added two structural guards: the rule is declared rather than hardcoded, and
every declared trigger must name a real argument of that tool, checked against
the published `inputSchema` — a typo in a trigger name would silently
un-promote the tool, which is the exact failure this prevents.

## Tests

| | |
|---|---|
| `test_grep_with_replace_is_refused_for_read_only_oauth_caller` | OAuth path |
| `test_grep_with_replace_is_refused_for_read_only_pat` | PAT path |
| `test_grep_without_replace_stays_readable_for_read_only_pat` | non-regression |
| `test_grep_with_replace_passes_when_write_scope_present` | positive |

The same misclassification existed in the audit trail (`_READ_ONLY_TOOLS`), so
a mutating call could be dropped from the audit stream when read-logging was
off. `record_tool` now takes `is_write` from the dispatch chokepoint, and a
test pins the two classifications against each other.

Verified locally against the docker stack across the read/write × with/without
matrix, plus audit behaviour with read-logging disabled.

### DIFF
--- a/backend/app/services/audit_log.py
+++ b/backend/app/services/audit_log.py
@@ -63,12 +63,23 @@ _GENESIS = "0" * 64
 # new write-tool can never silently escape the audit trail. (File tools
 # `akb_get_file`/`akb_put_file` are proxy-only and never reach this
 # backend dispatch, so they're intentionally absent.)
+#
+# Membership here must agree with the read half of `_TOOL_SCOPES` in
+# mcp_server.server — a tool that needs `akb:vault:write` is by definition
+# state-changing. A test asserts the two agree; `akb_publication_snapshot`
+# was listed here while mapped write-grade, so with `log_reads` off it
+# wrote an S3 snapshot and flipped a publication's mode with no audit line.
+#
+# Tool-level membership is not sufficient on its own: `akb_grep` is a read
+# tool whose `replace` argument rewrites every matching document. The
+# caller therefore passes `is_write`, computed by the same
+# `_required_scope` that gates the call, so an argument-driven write is
+# recorded even though the tool name sits in this set.
 _READ_ONLY_TOOLS = frozenset({
     "akb_get", "akb_search", "akb_browse", "akb_drill_down", "akb_grep",
     "akb_graph", "akb_relations", "akb_history", "akb_diff", "akb_activity",
     "akb_provenance", "akb_list_vaults", "akb_vault_info", "akb_vault_members",
     "akb_whoami", "akb_help", "akb_search_users", "akb_publications",
-    "akb_publication_snapshot",
 })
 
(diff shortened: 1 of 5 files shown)


kwoo24-oss:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #342, opened 2026-08-07
native grep replace still caps the rewrite at the output limit (legacy #315 fixed, native arm not) — cutover precondition


imjhyuk-oss:
## Summary

#315 is fixed for the legacy arm: `grep(replace=…)` now rewrites the whole
matched set, with `limit` demoted to a preview knob and a separate fail-closed
`grep_replace_max_docs` budget bounding the write. The **native arm has the same
defect and still has it**, so the two arms now disagree about what a replace
covers. Removing the M1 gates without closing this would make find-and-replace
silently narrower after cutover than it is today.

## Where

`backend/app/services/m1_native_grep_service.py`:

- `grep_public` → `grep` passes the caller's `limit` through as
  `resource_limit` into `_scan_bodies_sync` (`:674-719`).
- `_scan_bodies_sync` stops materialising resources at that cap
  (`:170`, `return_resource = mode == "default" and total_resources < resource_limit`)
  and records `truncation_reasons.add("resource_limit")` at `:251`.
- The replace path (`:740-834`) then writes `selected = matched` — i.e. the
  already-capped set.

So the native replace is bounded by the *output* `limit`, exactly as legacy was
before this fix. The native cap is `1..1000` (`:653`) rather than legacy's
50, which makes the truncation less likely to be hit, not less silent when it is.

## Why this is a cutover precondition, not a bug report

Neither arm is reachable in a stock deployment today — the native arm needs
`document_revision_backend = native_ledger_m1` (or a non-`s3_current` file
driver) **and** `native_revision_m1_measurement_only = true` **and**
`db_name = "akb_revision_m1_measurement"`. So nothing is broken for users right
now. The problem is directional: P2 removes those gates, and a user whose
find-and-replace works today would find it quietly truncating afterwards.

This is the same shape as #340 — a native-arm defect that is harmless while
gated and unacceptable the moment the gate comes off.

## Suggested direction

Mirror the legacy fix rather than reinventing it:

1. drive the replace from the full matched set, not the `resource_limit`-capped
   scan output;
2. reuse the same `grep_replace_max_docs` budget so both arms reject an
   over-wide scope at the same threshold, with the same `replace_scope_too_large`
   code;
3. keep `resource_limit` governing the returned snippet payload only, and make
   sure `truncation_reasons` cannot be read as "the rewrite was truncated".

Conformance case C7b pins the native grep's *read* semantics; it does not
currently pin replace scope, so closing this will want a C7b entry (or a new
[...]

Issue #341, opened 2026-08-07
grep replace: with regex=false the replacement string is literal when case_sensitive=true but a regex template when false (the default)


imjhyuk-oss:
## Summary

With `regex=false` the two case branches of `grep(replace=…)` disagree about what
the replacement string *is*. The case-sensitive branch treats it as literal text;
the case-insensitive branch — **the default** — passes it to `re.sub` as a regex
replacement *template*. So on the default path a replacement containing a
backslash is silently reinterpreted, and one containing `\1` raises an uncaught
`re.error` in the middle of a write loop that has already committed documents.

Both `regex` and `case_sensitive` default to `false`, so this is the path an
ordinary `akb_grep(pattern=…, replace=…)` takes.

## Where

`backend/app/services/search_service.py:1308-1315` (at `d8c1bee`):

```python
if regex:
    new_body = _re.sub(pattern, replace, body, flags=re_flags)
else:
    if case_sensitive:
        new_body = body.replace(pattern, replace)          # literal
    else:
        # Case-insensitive non-regex replace
        new_body = _re.sub(_re.escape(pattern), replace, body, flags=_re.IGNORECASE)
```

The `pattern` is correctly escaped on the third branch. The `replace` is not —
and `re.sub`'s second argument is a template, not a literal.

## Reproduction

Verbatim transcription of those eight lines, `regex=False`, body `"see TODO here"`,
pattern `"TODO"`:

| replacement | `case_sensitive=True` | `case_sensitive=False` (default) |
|---|---|---|
| `TODO\1` | `see TODO\1 here` | **raises `re.error: invalid group reference 1`** |
| `[\g<0>]` | `see [\g<0>] here` | `see [TODO] here` |
| `C:\new` | `see C:\new here` | `see C:` + newline + `ew here` |
| `DONE` | `see DONE here` | `see DONE here` |

Only the last row agrees.

## Why the exception matters more than it looks

`_re.sub` at `:1315` and `doc_service.update` at `:1325` are both outside the
`try` that guards `get` at `:1300-1304`. A `re.error` therefore propagates out of
`grep()` with no response object, so the caller learns neither that the run
aborted nor which documents were already committed — the same recovery gap
#315 describes for a mid-loop write failure.

In practice the template is a property of `(pattern, replace)` rather than of any
one document, so it fails on the first document that would have changed and the
blast radius is zero-or-one commits. That is luck, not design.

## Contract

[...]

Issue #340, opened 2026-08-07
native resolve_live_reference: a UUID that is both a Resource id and another Resource path silently changes target


imjhyuk-oss:
`NativeRevisionRepo.resolve_live_reference`
(`backend/app/repositories/native_revision_repo.py:170-195`) tries
`resource_id` first and, on a hit, returns immediately. The fall-through added
in #335 is correct and its comment is correct:

> Fall through to path/alias resolution — do NOT return None here. A
> UUID-shaped reference that does not name a live Resource by id may still name
> one by path.

That handles the id-**miss**. It does not address the id-**hit collision**: when
the same UUID string is both a live Resource's id *and* another live Resource's
path in the same namespace and surface, the id now wins where the path used to.

## Why the collision is reachable

The docstring's own premise is that File paths may be bare UUID-shaped strings
— `app/util/text.validate_file_name` imposes no extension rule. So in one
namespace/surface it is possible to have:

- `F1`, a live File whose `resource_id` is `U`;
- `F2`, a live File whose `current_path` is literally the string `U`.

## Reproduction

Same disposable PostgreSQL, base tree vs. HEAD tree:

```
ef8ee6d:  resolve_live_reference(<F1.id>)  ->  F2   (path match)
d8c1bee:  resolve_live_reference(<F1.id>)  ->  F1   (id match)
```

## Why it is not read-only

`resolve_live_reference` is reached through `_lock_live_reference`
(`backend/app/services/native_revision_service.py:274`, resolving at `:291` and
`:315`), which is the target selector for `replace_text`, `move_text` and
`delete_resource`. Consequences for the pair above:

- `F2` is no longer addressable by its own path;
- a delete aimed at `F2` by path selects and deletes `F1`.

## Scope

Native arm only. Both entry points are behind the M1 gates
(`document_revision_backend = native_ledger_m1`, or
`native_revision_m1_file_driver != s3_current`), each of which additionally
requires `native_revision_m1_measurement_only = true` and
`db_name = "akb_revision_m1_measurement"`. It therefore does **not** affect the
P1 exit condition — legacy remains the sole write authority — and it cannot
reach a stock deployment. It matters because P2 migration is what removes those
gates, and a resolution rule that silently changes its target is a bad thing to
carry across a cutover.

## Suggested direction

The two namespaces are genuinely distinct, so the fix is probably to stop
letting them overlap silently rather than to reorder the lookups:

1. resolve both, and treat a double hit as ambiguous (the spec already has an
[...]

Issue #315, opened 2026-07-28
akb_grep(replace=…) rewrites only the first `limit` matching documents, contradicting the documented "every matching document" behaviour


imjhyuk-oss:
## Summary

A find-and-replace across a scope with more matches than `limit` silently rewrites only the first `limit` documents and commits them, while the tool schema and help promise it rewrites *every* matching document. `limit` is documented as an output-only knob and is hard-capped at 50.

## Where

`SearchService.grep` computes `matched_docs` (all matches) then slices `result_docs = matched_docs[:limit]` (`backend/app/services/search_service.py:995-998`); the replace loop iterates only `result_docs` (`:1006`, `:1009`, commit at `:1039`). `limit` is clamped to `settings.search_limit_max` = **50** (`:844` → `:117`, `backend/app/config.py:204`), so more than 50 matches cannot be fully rewritten in any single call.

## Contract mismatch

The documentation says the opposite in several places:

- `backend/mcp_server/tools.py:435` — `replace` "replaces all matches in **every matching document across the search scope**"
- `backend/mcp_server/tools.py:436` — `limit` is "Max documents to return", `maximum: 50`
- `backend/mcp_server/tools.py:424-425` — advises switching to `count_only`/`files_with_matches` "instead of bumping `limit`", i.e. steers callers away from the only lever that would widen the write
- `backend/mcp_server/help.py:171`, `:859`; `backend/mcp_server/server.py:232-234` — "find-and-replace across all matching documents"

## Repeated calls do not always converge

Because matching is `ILIKE` / `pattern.lower() in line.lower()` (`:952`), two common cases never drain:

- **Case-only replacement in the default (case-insensitive) mode**, e.g. `postgresql` → `PostgreSQL`: after the first pass the doc still matches but `new_body == body`, so `:1032` `continue`s — the doc keeps occupying a slot and later matches are never reached. `replaced_docs=0 / total_docs=100` then reads to a caller as "nothing to do".
- **Replacement string contains the pattern**, e.g. `TODO` → `TODO(owner)`: the first `limit` docs are rewritten on every call (`TODO(owner)(owner)(owner)…`) while the rest are starved.

Selection is deterministic (`ORDER BY v.name, d.path, c.chunk_index` + dict insertion order), so the same first-`limit` documents are chosen each time.

## Recovery gap

[...]

PR #343, opened 2026-08-07
fix(grep): make replace write the document set the caller asked for (#338, #315, #341)

### LINKED ISSUE #338: grep replace rewrites documents outside the requested collection (unanchored, unescaped LIKE)
## Summary

`grep(replace=…)` rewrites documents the caller never scoped. The `collection` filter is compiled as an unanchored, unescaped `LIKE`, so a rewrite scoped to one collection also rewrites every sibling collection whose path shares that prefix — each with its own git commit and re-index.

This fires on ordinary values, not adversarial ones: `collection='core'` also matches `core-extra/…`.

## Where

`backend/app/services/search_service.py:1170` (and the sibling at `:461`):

```sql
d.path LIKE $n || '%'
```

No anchor, no `ESCAPE`. Confirmed in PostgreSQL:

- `'core-extra/b.md' LIKE 'core%'` → `t`
- `'cork/c.md' LIKE 'cor_%'` → `t`  (`_` acts as a single-char wildcard)
- `collection='%'` → matches every accessible document

## Why the write path is the real problem

On the read path the blast radius is bounded and uninteresting: the ACL clause and the vault clause are separate ANDed conditions, so no metacharacter in `collection` reaches a vault the caller cannot already read, and the worst outcome of `collection='%'` is what omitting the parameter already returns. Parameters are bound by asyncpg, so this is **not** SQL injection and **not** a confidentiality issue.

The write path is different. `grep(replace=…)` applies find-and-replace to every document in `result_docs`, which comes from that same over-matching `WHERE` (`search_service.py:1281-1294`). So a user scoping a rewrite to `collection='core'` silently also rewrites `core-extra/…` — documents they never named — producing a git commit plus a re-index per document. A `%` or `_` in the collection name widens that to the whole accessible vault.

Reachability: `collection` is a free-form, unvalidated string on both public surfaces — REST `GET /search/grep` (`api/routes/search.py:88`) and MCP `akb_grep` (`mcp_server/tools.py:437`, passed through unmodified at `server.py:758`).

## The fix already exists in this product

`backend/app/services/m1_native_grep_service.py:511-512` gets the containment right:

```sql
rs.current_path = $n OR rs.current_path LIKE $n ESCAPE '\'
```

Backport that shape to `search_service.py:461` and `:1170`. `backend/app/services/agent_memory_service.py:840,893` should be reviewed for the same pattern.

A regression test for the write path matters more than one for the read path: scope a `replace` to a collection that has a prefix-adjacent sibling and assert the sibling is untouched (no new commit, no re-index).

## Provenance

Found while making a downstream conformance case executable. The case pins today's legacy behaviour rather than promoting it to a golden — the harness is not blocked on this fix. Native grep already behaves correctly, so a companion case shows the anchored/escaped semantics side by side with the legacy quirk.

Severity assessed as Medium-High: scope containment with write amplification, triggered by honest input.

imjhyuk-oss:
Three defects, one function, one shape: **`grep(replace=…)` wrote the wrong
thing.** They were worth fixing together because two of them are the same loop
disagreeing about *which* documents, and the third decides *what bytes* go into
them — and widening the scope without fixing the bytes would have widened the
blast radius of an uncaught exception.

| | defect | direction of the error |
|---|---|---|
| **#338** | `collection` filter was an unanchored, unescaped `LIKE` | rewrote documents **outside** the requested scope |
| **#315** | replace iterated `matched_docs[:limit]`, `limit` capped at 50 | rewrote **only part** of the requested scope, silently |
| **#341** | non-regex replacement was literal on one case branch, a regex template on the other | wrote **the wrong bytes**, or raised mid-loop |

## #338 — scope containment

The filter compiled to `d.path LIKE $n || '%'`. Unescaped, so a `_` or `%` in a
perfectly ordinary collection name was a wildcard; unanchored, so `core` also
admitted the sibling collection `core-extra/`. Every swept-in document got its
own git commit and re-index.

The product already had the right rule in three places — the native arm
(`m1_native_grep_service._head_bodies`), the repo layer
(`document_repo.list_docs_under` and friends), and conformance case **C7b**,
which pins it as the *target* containment: `path == prefix or
path.startswith(prefix + "/")`. `search_service` was the outlier that never
adopted it. This adds one `collection_containment_sql` helper emitting

```sql
(<column> = $exact OR <column> LIKE $prefix ESCAPE '\')
```

over `like_escape(prefix) + "/%"`, and applies it at **all three** sites — the
grep write path plus the two read-path siblings at `:461` and `:541` that #338
also names. It replaces two hand-inlined copies of the escape triple-replace
with the existing `util.text.like_escape`.

The exact-match arm is inert on `documents.path` (a document path is always
composed by `doc_path()` and carries a `.md` segment) and load-bearing on
`collections.path` / `resources.current_path`. Kept everywhere so the three
sites read as one rule rather than three dialects — and so legacy does not
diverge from the contract C7b pins for P2.

## #315 — replace covers the scope, and fails closed

`limit` is documented as an output knob and clamped to 50, so driving the
rewrite from `result_docs` dropped every match past the 50th while reporting
success. Worse, it did not converge on retry: a case-only replacement
(`postgresql` → `PostgreSQL`) leaves the doc matching but unchanged, so the same
first 50 documents were re-selected forever.

Replace now covers the full matched set. `limit` governs the preview only, and
says so in the response hint when both are in play. The write gets its **own**
bound — `grep_replace_max_docs` (default 500) — which **fails closed**: a scope
wider than the budget is rejected with `code: "replace_scope_too_large"` and
**nothing written**, pointing the caller at `count_only` / `files_with_matches`.
Truncating a write silently is what caused this; truncating it loudly would too.

A partial run is now recoverable, which it was not: `doc_service.update` moved
inside the guard, every entry carries `previous_commit` as well as `commit`, and
an abort returns `code: "replace_incomplete"` with the ledger of what already
landed instead of letting the exception discard it.

## #341 — the replacement string is data, not a template

Found while reading the same eight lines. With `regex=false`, `case_sensitive=True`
used `str.replace` (literal) but the **default** `case_sensitive=False` passed the
string to `_re.sub`, making it a regex replacement template:

```
replacement    case_sensitive=True     case_sensitive=False (default)
TODO\1         literal                 raises re.error, uncaught, mid-write-loop
[\g<0>]        literal                 expands to the matched text
C:\new         literal                 writes a newline
```

A function replacement is inserted verbatim, so both branches now agree.
Backreferences remain a regex-mode feature (as documented), and a bad template
is now reported rather than escaping.

## Verification

```
backend unit  -k 'not _e2e'   1608 passed, 398 skipped, 36 deselected, 0 failed
                              (1582 at base; the delta is exactly this PR's 26)
ruff 0.15.16                  All checks passed
mypy 2.1.0 --python-version 3.14   Success: no issues found in 243 source files
bandit 1.9.4 medium+          0
```

**The mocked tests are not vacuous.** `_FakeConn` evaluates the collection
predicate the query actually emits, `ESCAPE` semantics and all, so
"the prefix-adjacent sibling is untouched" is asserted end-to-end against
`doc_service.update` call records rather than by eyeballing a SQL string. That
mini-`LIKE` was cross-checked against **PostgreSQL 16 on 12 cases, 0
disagreements**.

Two things a fake connection structurally cannot catch, so both were checked
against a real database:

- **Parameter drift.** The containment predicate binds *two* params where the
  old one bound one, so every downstream `$n` shifts. A fake conn accepts a
  query whose placeholders no longer match its params; PostgreSQL does not.
  Pinned by `test_every_placeholder_is_bound_with_all_filters_on`.
- **The SQL itself.** The exact SQL and params `grep()` builds were executed
  against a disposable PostgreSQL 16 with a real schema and a corpus containing
  `core/`, `core/deep/`, `core-extra/`, `my_docs/`, `myXdocs/` and a root
  `coreutils.md` — **6 scenarios, 0 mismatches**, including `collection="%"` and
  `collection="cor_"` now returning nothing instead of the whole vault.

**Not run:** the 15-suite shell E2E gate (needs a live backend, MinIO and an
embedding stub). `test_graph_replace_e2e.sh:169-198` is the tripwire there — it
asserts `replaced_docs == 3` on 3 documents and `≥ 1` on a regex+collection
call, both far under the budget and inside a collection with no prefix-adjacent
sibling, so it should be unaffected.

## Harness impact: none by default

Conformance case **C7a** records this exact bug as `C7A-D02` with disposition
`record_projection_only`. Its default offline run replays a frozen transcript
against a stdlib oracle that deliberately reimplements the buggy `sql_like()` —
it never calls the product, so `run.sh --case C7a` keeps passing unchanged. Only
an explicit live re-run through `AKB_NATIVE_REVISION_C7A_TRANSCRIPT` against a
fixed AKB would fail, and it fails with the designed message
`"C7A-D02 is no longer exercised…"` — the signal to retire the defect record.
That retirement is a workbench change, not part of this PR.

## Registered, not fixed

- **#342** — the native arm caps its replace at `resource_limit` exactly as
  legacy did. Filed rather than fixed here: it is gated measurement code the P1
  exit audit just cleared at `d8c1bee`, and widening this PR into it would mean
  re-establishing the C7b evidence in the same change. It is a **cutover
  precondition**, same shape as #340.
- `agent_memory_service` builds LIKE patterns from `sanitise_agent_id` /
  `sanitise_session_id` output, which deliberately preserves `_`, unescaped.
  Read-only and confined to the caller's own private vault, so at most a
  same-user cross-session lookup ambiguity — recorded, not filed.
- The audit line for a grep replace still records no `pattern`, no `replace`,
  and no `replaced_docs` (`audit_log.py:372-380`; `pattern` is not in
  `_TARGET_KEYS`). That gap matters more now that a single call can touch up to
  the budget, but extending the audit payload is a product decision, not a
  bug fix.

Closes #338, #315, #341.

https://claude.ai/code/session_019rcTyzcrzMgCN7pxArJZqk

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -158,7 +158,7 @@
         "filename": "backend/mcp_server/help.py",
         "hashed_secret": "e5e9fa1ba31ecd1ae84f75caaa474f3a663f05f4",
         "is_verified": false,
-        "line_number": 357,
+        "line_number": 358,
         "is_secret": false
       },
       {
@@ -166,7 +166,7 @@
         "filename": "backend/mcp_server/help.py",
         "hashed_secret": "90bd1b48e958257948487b90bee080ba5ed00caa",
         "is_verified": false,
-        "line_number": 1151,
+        "line_number": 1165,
         "is_secret": false
       }
     ],
@@ -5372,5 +5372,5 @@
       }
     ]
   },
-  "generated_at": "2026-08-05T05:14:27Z"
+  "generated_at": "2026-08-07T13:27:48Z"
 }
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -484,6 +484,18 @@ class Settings(BaseModel):
     # caller (MCP, REST, internal) is bounded uniformly.
     search_limit_max: int = Field(default=50, ge=1)
 
+    # Hard ceiling on how many documents a single `grep(replace=...)` may
+    # rewrite. Deliberately NOT `search_limit_max`: that one bounds the
+    # snippet payload, and reusing it as the write bound is what made a
+    # find-and-replace silently rewrite only the first 50 matches and report
+    # success (issue #315). `limit` is now preview-only; this is the write
+    # bound, and it fails CLOSED — a replace whose scope exceeds it is
+    # rejected with zero writes rather than truncated, so the caller narrows
+    # the scope deliberately instead of discovering a partial rewrite later.
+    # Each replaced document costs one git commit plus one re-index, which is
+    # what keeps this a bounded number rather than "all of them".
+    grep_replace_max_docs: int = Field(default=500, ge=1)
+
     # Push the ACL filter down to VAULT granularity in the vector store (issue
     # #189 Phase 2). When True AND the driver is pgvector AND a search has no
     # doc-level filter (collection/doc_type/tags/source_uris), search filters by
--- a/backend/app/models/document.py
+++ b/backend/app/models/document.py
@@ -323,6 +323,10 @@ class GrepReplacement(BaseModel):
     path: str | None = None
     title: str | None = None
     commit: str | None = None
+    # The commit the document moved OFF. A replace is one commit per
+    # document with no transaction spanning them, so a run that aborts
+    # part-way is only unwindable if each entry carries both ends.
+    previous_commit: str | None = None
     error: str | None = None
 
 
--- a/backend/app/services/search_service.py
+++ b/backend/app/services/search_service.py
@@ -27,6 +27,7 @@ from app.services.vector_store import VectorHit, get_vector_store
 from app.services.vector_store.base import VectorStoreUnavailable, supports_vault_filter
 from app.services.rerank_service import RerankError, rerank
 from app.services.uri_service import parse_uri
+from app.util.text import like_escape
 
 logger = logging.getLogger("akb.search")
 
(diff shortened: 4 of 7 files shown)


imjhyuk-oss:
Superseded. All three defects this branch targeted are fixed on `main`, by
other changes that landed while it sat:

| issue | fixed by |
| --- | --- |
| #315 — `limit` truncated the replace scope | #348 |
| #338 — unanchored, unescaped collection `LIKE` | #569 |
| #341 — literal replacement treated as a regex template | #569 |

The coverage landed with them:

- `backend/tests/test_grep_replace_unit.py` — `limit` bounds the preview and
  not the scope; a literal replacement keeps its backslashes.
- `backend/tests/test_search_filter_contract_unit.py::test_collection_boundary_and_literal_metacharacters`
  — the collection predicate is anchored and its metacharacters are literal.
- `app/services/search_filters.py` now owns `escape_like` and
  `collection_predicate`, so the escaping is a named function rather than a
  call site that has to remember.

#569 also reworked the grep path itself (30 files), so this branch is a
rewrite against code that no longer has this shape, not a rebase. Closing it
rather than carrying it — the branch stays in history if any of its test cases
are worth lifting later.

PR #447, opened 2026-08-31
feat(sso): offer this installation own realm as a login option

### LINKED ISSUE #446: An sso installation with no external IdP cannot be entered at all
## An `sso` installation with no external IdP cannot be entered at all

A workspace running `auth_mode: sso` with its own bundled Keycloak, and no
external identity provider, has no way for a person to sign in. Not a degraded
way — none. This is reachable with nothing unusual: provision an SSO
installation, do not register an upstream, and it is already in this state.

It is a regression against the provider-less path that `/auth/keycloak/login`
used to serve, and the replacement covers only the brokered case.

### Reproduction

A freshly provisioned SSO installation reports a healthy contract:

```json
{"schema_version": 2, "auth_mode": "sso",
 "local_auth": {"enabled": false},
 "keycloak": {"enabled": true, "browser_session_ready": true},
 "providers": []}
```

`providers` is empty, so no `login_url` exists. The retired route answers:

```
GET /api/v1/auth/keycloak/login
410  "Select one of the enabled SSO providers from /api/v1/auth/config"
```

There is nothing to select.

### Why the two obvious ways out are both closed

**Register the installation's own realm as its provider** — refused, correctly:

```python
# backend/app/sso/providers/keycloak_oidc.py
if _issuer_identity(issuer) == _issuer_identity(broker_issuer):
    raise ProviderDefinitionError("provider_issuer_is_broker")
```

That guard is right. `begin_browser_login` attaches `kc_idp_hint` unconditionally
and provider registration creates a broker identity provider inside the realm, so
a self-referential provider is an authorize loop, not a login. The guard refuses a
broken expression of the intent; it is not the cause of this issue.

**Add a second realm on the same Keycloak** — not reachable. The bootstrap
administrator is retired as part of a completed install (recorded in
`standalone_sso_bootstrap_retirements`), which is a deliberate and good property;
the consequence is that no realm-creating credential survives the install.

So the only remaining options are to obtain an external IdP or to broker to some
other party's realm. For a self-contained deployment — an evaluation, an air-gapped
or single-team install, a first-run before any corporate IdP exists — neither is
available, and the installation is permanently uninhabitable.

### What is actually missing

Not a relaxation of the guard. A provider **kind** that names the installation's
own realm as the place people authenticate, so that gen-2's improvement — the
caller names which upstream, rather than there being an implicit one — still
holds, while a workspace with no upstream is expressible.

Three places assume brokered, and each needs the distinction rather than a
loosening:

| where | today | needed |
| --- | --- | --- |
| `begin_browser_login` | `kc_idp_hint` always attached | omitted for the local kind |
| callback claim check | `claims["identity_provider"]` required and must equal the selected alias | for the local kind: **must be absent** |
| browser session service | `identity_provider` a required claim | same distinction |

The middle row is the load-bearing one. `identity_provider` must not become
optional — an optional claim is a hole, and a token from one provider could then
be presented for another. The kind decides which assertion applies: present and
matching for a brokered provider, **absent** for the local one. Absence asserted,
never tolerated.

`backend/app/sso/registry.py` is already an explicit registry with a
`ProviderDefinition` per type, so the kind is an addition there rather than a
special case threaded through call sites.

### The cost, stated

The installation's own realm also holds the product administrator account, so
enabling this kind puts the operator plane and the audience plane in one realm.
That is a real trade and should be the workspace owner's explicit choice rather
than a default. It is bounded rather than open: an administrator who signs in
this way still arrives as a pending admission and still needs approval, exactly
like anyone else, so arrival is not entry.

### Acceptance

- An SSO installation with no external IdP can be configured so that a person
  signs in with an account in its own realm, and appears as a normal pending
  admission.
- A brokered provider still requires `identity_provider` present and matching;
  the local kind requires it absent. Neither accepts the other's shape.
- No self-referential broker provider can be registered: `provider_issuer_is_broker`
  stays exactly as it is.
- The kind is off unless configured, so an installation that registers no
  provider keeps today's behaviour rather than silently gaining a login.

imjhyuk-oss:
Closes #446.

An `sso` installation with its own bundled Keycloak and **no external identity
provider has no way in.** Not a degraded way — none. `providers` is empty so no
`login_url` exists, the retired provider-less route answers 410 with "select one
of the enabled SSO providers", and there is nothing to select.

The two obvious ways out are both closed, and correctly:

| | |
| --- | --- |
| register the realm as its own provider | `provider_issuer_is_broker` |
| add a second realm on the same Keycloak | no realm-creating credential survives a completed install |

**The first guard is right and is untouched here.** `begin_browser_login`
attaches `kc_idp_hint` unconditionally and registration creates a broker identity
provider inside the realm, so a self-referential provider is an authorize loop,
not a login. A test pins it so this change cannot be read as relaxing it.

## The claim does not become optional

That is the whole design. A brokered token carries `identity_provider`; a direct
realm sign-in cannot, because nothing brokered it. So the **kind** decides which
assertion applies — present and matching for a brokered provider, **absent** for
the local one — and each rejects the shape the other produces. An optional claim
would be a hole: a token minted through one provider could then be presented for
another.

The alias is reserved; both provider modules refuse it, because the discrimination
rests on the two kinds never sharing a name.

## Where it lives, and why not the registry

The provider catalog **is** Keycloak's identity-provider list — `_fetch_catalog`
reads it, and there is no provider table on this side — so a kind that
materialises nothing has nowhere to live in that store. It is a setting,
projected into the catalog at the one place that catalog becomes a browser's list
of choices, and answered *before* the catalog is consulted on the login path, so
a broker outage cannot take away the one login that does not depend on a broker.

Off by default: an installation that registers no provider keeps today's
behaviour rather than silently gaining a login.

## The cost, in the setting rather than hidden

This realm also holds the product administrator account, so enabling the kind
puts the operator plane and the audience plane in one realm. Bounded, not open:
an administrator who signs in this way still arrives as a pending admission and
still needs approval, so arrival is not entry.

## Verification

12 new tests. Three mutations, each red on its own assertion and green again on
restore:

| mutation | red |
| --- | --- |
| remove the biconditional | 2 |
| remove the reserved-alias guard | 1 |
| remove the absence mapping in the session service | 1 |

Regression as a side-by-side control **in one container**, same install, same
selector:

```
origin/main    521 passed, 95 skipped, 0 failed
this branch    533 passed, 95 skipped, 0 failed
```

The difference is exactly the 12 added. An earlier run of that comparison
reported 12 failures on both sides and was not evidence: only `backend/` was
mounted, so conftest could not reach `config/app.yaml.example` at the repo root.

`scripts/check.sh`: all checks passed.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "backend/CHANGELOG.md",
         "hashed_secret": "5315196a87449ee7110cd2cce3695bdcf505e55f",
         "is_verified": false,
-        "line_number": 3634
+        "line_number": 3664
       }
     ],
     "backend/mcp_server/help.py": [
@@ -5400,5 +5400,5 @@
       }
     ]
   },
-  "generated_at": "2026-08-31T06:01:05Z"
+  "generated_at": "2026-08-31T10:55:38Z"
 }
--- a/backend/app/api/routes/auth.py
+++ b/backend/app/api/routes/auth.py
@@ -46,6 +46,7 @@ from app.sso.keycloak_admin import (
     get_keycloak_provider_control,
 )
 from app.sso.providers.keycloak_oidc import ProviderDefinitionError, validate_alias
+from app.sso import local_realm
 from app.util.text import NFCModel
 
 router = APIRouter()
@@ -189,6 +190,21 @@ async def auth_config(response: Response):
             for provider in catalog
             if provider.state == "enabled"
         ]
+        # The installation's own realm, when its owner asked for it. It is not a
+        # Keycloak identity provider, so it is not in the catalog read above; it
+        # is appended here, at the one place that catalog becomes a browser's
+        # list of choices. Its alias is reserved, so it cannot collide with one.
+        if settings.sso_local_realm_login_enabled:
+            providers.append(
+                {
+                    "provider_type": local_realm.PROVIDER_TYPE,
+                    "alias": local_realm.ALIAS,
+                    "display_name": settings.sso_local_realm_display_name,
+                    "login_url": (
+                        f"/api/v1/auth/sso/{local_realm.ALIAS}/login" if browser_session_ready else None
+                    ),
+                }
+            )
     return {
         "schema_version": 2,
         "auth_mode": mode,
@@ -347,6 +363,16 @@ async def _require_enabled_provider(alias: str) -> str:
             status.HTTP_404_NOT_FOUND,
             "SSO provider is not enabled",
         ) from None
+    # The local realm is answered before the catalog is consulted: it is not in
+    # there, and a broker outage must not take away the one login that does not
+    # depend on a broker.
+    if local_realm.is_local_alias(alias):
+        if not settings.sso_local_realm_login_enabled:
+            raise HTTPException(
+                status.HTTP_404_NOT_FOUND,
+                "SSO provider is not enabled",
+            )
+        return alias
     control = get_keycloak_provider_control()
     if control.control_mode != "direct":
         raise HTTPException(
@@ -373,6 +399,15 @@ def _require_signed_provider(
     expected_alias: str,
 ) -> None:
     value = claims.get("identity_provider")
+    # Present and matching for a brokered provider; ABSENT for the local realm,
+    # which brokers nowhere and so is never vouched for by anyone. The claim does
+    # not become optional -- an optional claim is a hole, and a token minted
+    # through one provider could then be presented for another. Each kind asserts
+    # the shape the other cannot produce.
+    if local_realm.is_local_alias(expected_alias):
+        if value is not None:
+            raise AuthenticationError("SSO sign-in failed")
+        return
     if not isinstance(value, str):
         raise AuthenticationError("SSO sign-in failed")
     try:
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -930,6 +930,18 @@ class Settings(BaseModel):
     # user plus exact binding. `invite_only` accepts only an exact prebound
     # (issuer, subject) identity. `disabled` rejects external login entirely.
     keycloak_enrollment_mode: Literal["open", "invite_only", "disabled"] = "open"
+    # Offer the installation's OWN realm as a login option, for a deployment
+    # that has no external identity provider to broker to. Off by default: an
+    # installation that registers no provider keeps today's behaviour rather than
+    # silently gaining a login.
+    #
+    # The cost is real and is the operator's to accept: this realm also holds the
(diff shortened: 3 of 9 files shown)


imjhyuk-oss:
**Live verification against a real workspace — and it found a defect the unit
suite could not see.**

A workspace was provisioned with its own bundled Keycloak, `auth_mode: sso`, and
no external identity provider: the exact state #446 describes. It served
`providers: []` and had no way in.

With this branch built and the setting enabled, the public contract gains the
kind and the login route behaves:

```
/api/v1/auth/config     providers: [{provider_type: "local-realm", alias: "local",
                                     login_url: "/api/v1/auth/sso/local/login"}]
/auth/sso/local/login   303 -> the workspace own realm authorize, NO kc_idp_hint
/auth/sso/nope/login    404
/auth/keycloak/login    410   (the retirement is untouched)
```

## The defect

The first live sign-in answered **401 at the callback**. The route's check was
patched and `verify_browser_id_token` was not — it carried its own
`identity_provider` requirement, unconditionally, so a local-realm token was
rejected one layer below the assertion that was supposed to decide.

The unit suite could not have caught it as written: it asserted the route helper,
and nothing reached the service. Fixed by applying the same biconditional there,
and **two tests were added for that site specifically** before rebuilding.

## A person actually signed in

Driven through a real browser, after the fix:

```
ok  the browser lands on this workspace own realm
ok  no broker hint is carried
ok  the workspace issues its __Host- session cookie
ok  /auth/me answers 200 to that browser session
    {"username":"admin", ..., "auth_method":"bro...
4 passed, 0 failed
```

## Negative controls

```
brokered workspaces unchanged   two workspaces on a `platform` broker still report providers=['platform']
the switch actually switches    setting it false -> providers: [] and /auth/sso/local/login -> 404
```

Tests are now **14 passed**. Everything else in the description above still
holds: three mutations red on their own assertions, and the side-by-side
regression control (origin/main 521 / this branch 533, zero failures on either).

imjhyuk-oss:
**Correction to "The cost, in the setting rather than hidden" above.**

That section, and the code comment it shipped, said enabling the kind "puts the
operator plane and the audience plane in one realm". It does not. A realm-local
native identity is already what this realm is for — the product administrator is
one — and in `sso` mode this realm is the only human issuer the installation
accepts, so a brokered person's account lives here too, bound to an upstream.
The setting decides something narrower: whether ordinary people hold a native
account here rather than one bound to an upstream. Off by default is still
right; the reason given for it was not.

Full reasoning in dnotitia/akb#446, correction in dnotitia/akb#449, which also
closes the documentation gap that made this feature unreachable to the
standalone installation it was written for.

PR #468, opened 2026-09-02
feat(helm): automate bundled secret manager bootstrap

kwoo24-oss:
## Summary

- make a single Helm install complete bundled OpenBao/Vault initialization, first unseal, policy/KV bootstrap, and Vault Secrets Operator projection
- add a namespace-scoped bootstrap Job with resumable, idempotent receipts
- retain native recovery material in an explicit one-time Kubernetes Secret handoff instead of Helm values, release metadata, or logs
- enable the installation-owned Keycloak realm login provider in standalone SSO profiles
- document recovery custody, PGP, Auto Seal, upgrades, and the shared VSO prerequisite

## Motivation

The bundled Secret Manager profiles previously required a two-phase workflow: install a sealed server, run a host-side initializer, and then apply Helm again. That made plain Helm installs incomplete and split lifecycle ownership between the chart and a shell orchestrator.

This change makes the chart the lifecycle boundary. After cluster prerequisites are present, `helm upgrade --install --wait --wait-for-jobs` waits until Secret Manager bootstrap, secret projection, SSO, and the application are ready.

## How it works

The chart creates a normal Kubernetes Job, rather than a Helm hook. The Job:

1. performs the engine's native initialization
2. stores native recovery output in `Secret/akb-secret-manager-recovery`
3. performs the first unseal for plaintext Shamir mode
4. waits for an active Raft member
5. configures KV v2, Kubernetes auth, runtime/operator policies, and AKB Secret Contract v1
6. revokes the initial root token and removes the transient bootstrap token
7. records a non-secret completion receipt
8. waits for every required VSO-projected Secret

A hash in the Job name creates a new Job when its specification changes. The receipt prevents a Helm upgrade from reinitializing an existing store.

## Recovery and seal behavior

- **Plaintext Shamir:** the first unseal is automatic. Native unseal shares are handed off once through the retained recovery Secret. Operators must copy them to approved off-cluster custody, verify the copy, and delete the Kubernetes Secret.
- **PGP:** one public key per share and a root-token public key are required at render time. Encrypted shares are never interpreted as plaintext credentials.
- **Auto Seal:** the Job waits for the configured KMS/HSM/Transit seal provider. Provider identity and key lifecycle remain operator-owned.

The recovery Secret uses `helm.sh/resource-policy=keep` so uninstall cannot silently destroy material that has not been handed off. Helm NOTES show the exact retrieval and deletion commands.

## Security boundaries

- generated values are not placed in Helm values, rendered manifests, release metadata, or Job logs
- bootstrap RBAC is namespace-scoped and restricted to named Secrets and ConfigMaps
- application workloads receive projected runtime credentials, not Kubernetes API permissions or root tokens
- the initial root token is revoked after bootstrap
- the backend remains independent of Vault/OpenBao APIs

VSO remains a cluster prerequisite because one controller can reconcile namespace-local connections and authentication resources for multiple releases. The application chart does not install or mutate cluster-scoped VSO resources.

## Review hardening

A second review and isolated upgrade rehearsal added the following safeguards:

- bind bootstrap receipt v3 to the actual Secret Manager cluster ID and immutable initialization/auth settings
- retry and read-after-write verify the recovery handoff before allowing bootstrap to continue
- predeclare an empty recovery Secret and receipt ConfigMap so the Job needs no namespace-wide create permission
- split Secret RBAC by operation: patch only the recovery handoff, delete only key-holder input, and read only projected runtime Secrets
- run the bootstrap container as UID/GID 1000 with a read-only root filesystem, dropped capabilities, and RuntimeDefault seccomp
- include every immutable Pod input and the chart version in the Job-name hash
- keep Helm-owned fields separate from Job-owned bootstrap status to avoid Server-Side Apply conflicts on upgrade

The tightened chart was installed in a new namespace, upgraded repeatedly through revision 5, and rechecked through the full SSO required-password-update and OIDC callback flow. The bootstrap identity was also verified to reject arbitrary Secret creation and mutation.

## Validation

Rebased onto the current `main` and verified with:

- `bash scripts/check.sh`
  - Ruff, mypy, Bandit, TypeScript, SDK/proxy contracts, frontend tests, and detect-secrets passed
  - frontend: 105 files / 732 tests passed
- `uv run pytest -q tests/test_secret_manager_bootstrap_unit.py tests/test_helm_deployment_unit.py tests/test_secret_management_deployment_unit.py`
  - 43 passed
- `helm lint` with the combined standalone SSO + Secret Manager profile
- fresh isolated Kubernetes deployment using bundled OpenBao and standalone Keycloak SSO
- repeated Helm upgrade without reinitialization
- browser redirect to the installation-owned Keycloak login form
- OIDC callback, browser session, CSRF-bound PAT creation, MCP initialization, vault/document CRUD, cleanup, and unauthenticated rejection

The live deployment exercised OpenBao + plaintext Shamir. HashiCorp Vault, PGP, and Auto Seal paths are covered by schema/render/unit validation but require their respective license acceptance, key holders, or external seal provider for live validation.

## Operator impact

Bundled profiles now finish with one Helm lifecycle. After a plaintext Shamir installation, operators must follow the Helm NOTES to transfer recovery shares off-cluster and delete the one-time recovery Secret. Future unattended restarts still require Auto Seal; automatic first unseal does not change Shamir's long-term custody model.

## Checklist

- [x] CI-equivalent static and contract checks pass
- [x] focused bootstrap/deployment tests pass
- [x] Helm lint passes
- [x] isolated deployment, upgrade, login, and CRUD scenarios pass
- [x] no credentials, private endpoints, or environment-specific deployment values are included
- [x] user-facing deployment and recovery behavior is documented

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "backend/CHANGELOG.md",
         "hashed_secret": "5315196a87449ee7110cd2cce3695bdcf505e55f",
         "is_verified": false,
-        "line_number": 3683
+        "line_number": 3700
       }
     ],
     "backend/mcp_server/help.py": [
@@ -6987,5 +6987,5 @@
       }
     ]
   },
-  "generated_at": "2026-09-02T08:06:11Z"
+  "generated_at": "2026-09-02T09:41:23Z"
 }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -149,7 +149,7 @@
         "filename": "backend/CHANGELOG.md",
         "hashed_secret": "5315196a87449ee7110cd2cce3695bdcf505e55f",
         "is_verified": false,
-        "line_number": 3700
+        "line_number": 3704
       }
     ],
     "backend/mcp_server/help.py": [
@@ -6987,5 +6987,5 @@
       }
     ]
   },
-  "generated_at": "2026-09-02T09:41:23Z"
+  "generated_at": "2026-09-02T10:51:01Z"
 }
--- a/README.md
+++ b/README.md
@@ -576,9 +576,9 @@ Kustomize-based profile installers.
   [`deploy/k8s/profiles/*/deploy.sh`](./deploy/k8s/profiles/README.md) when the
   installer should coordinate the complete lifecycle, including Secret
   Manager initialization where selected.
-- Use raw Helm or rendered Kustomize YAML only when the required Secrets,
-  cluster prerequisites, and any native Vault/OpenBao init/unseal/bootstrap
-  steps are already managed separately.
+- Raw Helm supports the complete bundled lifecycle when a Secret Manager
+  profile is selected and `--wait --wait-for-jobs` is used. Rendered Kustomize
+  YAML still expects its Secret Manager lifecycle to be managed separately.
 - Vault Secrets Operator is a cluster-scoped prerequisite. Its ownership and
   guarded removal are documented under
   [`deploy/helm/akb-cluster`](./deploy/helm/akb-cluster/README.md).
--- a/backend/CHANGELOG.md
+++ b/backend/CHANGELOG.md
@@ -19,6 +19,10 @@ Kubernetes Secret that operators must copy to off-cluster custody and delete;
 generated shares and tokens are never Helm values or release metadata. PGP
 mode keeps shares/root authority encrypted and waits for key-holder input,
 while Auto Seal continues to depend on its configured KMS/HSM/Transit provider.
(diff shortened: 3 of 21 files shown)


kwoo24-oss:
AKB가 Vault/OpenBao/VSO를 번들로 설치·초기화하는 방향을 철회했습니다. 애플리케이션은 운영자가 제공한 Kubernetes Secret만 소비하는 단순한 경계로 되돌리고 있으므로, 번들 secret manager의 Helm 자동 bootstrap을 추가하는 이 PR은 더 이상 적용하지 않습니다.

PR #496, opened 2026-09-07
perf(search): reduce pgvector reads within existing memory

kwoo24-oss:
## Summary

Reduce recurring pgvector/BM25 reads within the deployment's existing memory
budget. This PR does not introduce memory-size profiles, automatic prewarming,
search-timeout overrides, BM25 term pruning, candidate-limit changes, or a new
search engine.

- Add covering indexes for posting weights and chunk scope lookups on fresh
  posting-shaped stores.
- Provide an explicit concurrent-index maintenance script for existing stores;
  application startup never rebuilds a large existing index.
- Materialize an already-authorized source-ID candidate set before joining
  common BM25 posting lists. Wide vault filters retain the direct join so a
  large scope is not forced into memory.
- Treat explicitly empty authorized scopes as no matches before opening a
  database connection.
- Add privacy-safe service/driver stage timings for locating future bottlenecks.
- Add reproducible synthetic benchmarks and real-pgvector regression coverage.

## Why

At approximately one million vectors and 71.5 million posting rows, the
baseline sparse query repeatedly read posting weights from the heap and looked
up scoped chunks. Those reads consumed the same PostgreSQL/OS cache used by the
HNSW dense path. This made BM25 expensive by itself and caused a previously warm
dense path to read pages again after a mixed workload.

The change reduces the recurring working set instead of assuming that every
installation can increase memory. It preserves embeddings, BM25 weights, RRF,
reranking, result limits, and authorization.

## Measured result

The scale fixture used 997,935 synthetic 1,024-dimensional vectors, exactly
71,536,566 postings, an 8 GiB PostgreSQL container, and 2 GiB
`shared_buffers`. No production resource or setting was changed.

| Measurement | Before | After | Scope |
| --- | ---: | ---: | --- |
| Warm BM25 | 5,554 ms | 743 ms | 10% of vaults |
| Warm BM25 | 5,888 ms | 1,142 ms | 100% of vaults |
| Dense HNSW after mixed BM25/dense workload | 1,481 ms | 23 ms | 10% of vaults |
| First HNSW after PostgreSQL restart | 2,218 ms | 1,983 ms | 10% of vaults |

The result supports the recurring-read/cache-competition fix. It does **not**
claim that first-read storage latency is solved. The two added scale-test
indexes occupied 3.35 GiB while the old indexes remained, so disk/write cost
must be reviewed before existing-store maintenance.

## Safety and compatibility

- No application or Helm memory policy is added or changed.
- No automatic `pg_prewarm`, readiness delay, timeout expansion, or relevance
  heuristic is present.
- Existing indexes are preserved at startup. Operators opt into the concurrent
  maintenance SQL after checking disk, WAL/replica capacity, and write load.
- Empty scopes fail closed; logs contain timing/count metadata but no query,
  UUID, content, or credentials.
- Synthetic before/after checks preserve ordered IDs and exact scores in all 18
  benchmark cases. The source-ID branch is also exercised with and without a
  resource-type filter.

## Verification

- [x] 104 focused search regression tests passed.
- [x] 2 disposable real PostgreSQL + pgvector tests passed.
- [x] 1,000-chunk / 64,000-posting smoke benchmark passed all 18 rank-parity
  cases.
- [x] Ruff passed for changed Python files.
- [x] Mypy passed for both changed service modules.
- [x] Branch updated to the latest `main` used for this review.

See `docs/designs/search-performance.md` for reproduction details, limitations,
and the explicit maintenance procedure.


kwoo24-oss (commented):
## Automated self-review

Reviewed head `1c88da7322624bcfb47004bb444a645b43c30a1b` against `main`. This is an automated author-side review, not independent maintainer approval.

**No blocking findings remain in the reviewed changes.**

### Adjustments made during review

- Move empty-scope returns ahead of database initialization; add DB-free regression coverage so an empty authorized set cannot initiate work or widen the scope.
- Mark retrieval successful only after payload conversion succeeds, avoiding misleading success diagnostics on conversion failures.
- Run the new real-pgvector regression suite in the existing CI database job, rather than letting it skip in both CI jobs.

### Checked

- Covering columns do not change BM25/RRF/dense scoring or result-limit policy.
- Existing index definitions are preserved at startup. Maintenance remains an explicit operator action; concurrent creation and invalid-index inspection are documented.
- Source/vault filtering and no-access behavior remain fail-closed. New logs contain stage durations/counts, not raw query text, source IDs, bodies or credentials.
- Synthetic measurements distinguish warm queries, PG-cold/OS-unknown queries and mixed-workload eviction; the PR does not claim a production HTTP speedup or a complete cold-start fix.
- Local targeted units: 66 passed. Real-pgvector tests: four passed after review adjustments. Backend Ruff, mypy (355 source files), Bandit and staged-file secret checks passed. The earlier isolated HTTP scenario and scale experiments are documented with their limits.

### Non-blocking operational follow-ups

- First-query cache loading remains; bounded hot-block restoration needs a separate experiment.
- Wide-scope chunk lookups/aggregation, continuous writes, visibility-map churn and production query diversity are not solved by this PR.
- Existing stores require a separately approved maintenance window and disk/WAL review; no production changes were performed.

GitHub CI was still running when this review was posted; local checks are not a substitute for its final result.

kwoo24-oss (commented):
## Updated automated self-review

Reviewed head `ef3b806` against `main`. This is an author-side automated review, not independent maintainer approval.

**No blocking findings remain.**

### Review focus

- Confirmed the default configuration is unchanged: startup prewarming is `off` and hybrid document-frequency pruning is disabled at `0`.
- Confirmed source and vault filters remain mutually exclusive and authorization stays fail-closed. The durable backfill readiness check falls back to the existing source-ID path on errors.
- Confirmed source-candidate materialization changes join order only; real-pgvector regression coverage preserves ordered IDs and scores.
- Confirmed common-term pruning runs only when a dense embedding is available, preserves exact BM25 for sparse-only/degraded embedding paths, retains rare terms, and falls back to all terms when statistics lookup fails.
- Confirmed startup prewarming does not expose relation names, queries, credentials or document data in health metadata; it is serialized across API replicas and refuses readiness when the selected working set exceeds 90% of `shared_buffers`.
- Confirmed the million-vector Helm file is a composable tuning overlay for both public profiles, not a new deployment shape or a global default.
- Confirmed the reference resource limit accounts for cgroup page cache: the tested 24 GiB limit was rejected after reaching its ceiling; the 36 GiB profile completed at about 32.4 GB with no OOM event.
- Confirmed the synthetic benchmark limitations and the need for corpus-specific relevance, write-load and external-model validation are documented.

### Verification

- Full backend suite excluding descriptor-driven MCP integration jobs: 2,871 passed, 717 skipped, 0 failed.
- Targeted search/lifecycle/backfill/real-pgvector/Helm tests: 69 passed.
- Helm lint and standalone/standalone-sso render checks passed.
- Isolated Kubernetes restart/login/filter/concurrency/authorization/OOM scenarios passed; production was not modified and the test namespace was restored to baseline.
- GitHub CI: static analysis, backend unit, live pgvector, frontend E2E and repository-runtime shell E2E all passed.

### Remaining non-blocking operational limits

- The 16 GB cache / 36 GiB container values are a measured reference for the synthetic million-vector fixture, not automatic sizing.
- The 0.8 common-term cutoff is a latency/relevance policy choice and requires a real-query relevance gate before rollout.
- The first request still includes lazy tokenizer-process initialization (2.66 seconds in the restart scenario); a future optimization could initialize that worker during readiness if a lower restart ceiling is required.

PR #517, opened 2026-09-10
fix(cutover): let a File step aside instead of failing the whole cutover

imjhyuk-oss:
## What breaks today

`_apply_files` publishes every File at the `logical_path` the plan froze, and
lets `_publish_create` refuse when that path is already live. One taken path
ends the whole apply — including a run where the entire document inventory had
already migrated successfully.

## Why the collision is legitimate

Legacy File names carry no uniqueness guarantee:

- `UNIQUE(vault_id, s3_key)` covers the **storage key**, not the name.
- The partial unique index on `(vault, collection, name, content_hash)` does not
  apply to the historical upload path at all. Without a caller-supplied
  `content_hash` the key is random, which `file_service` documents as
  "re-uploading the same artifact always makes a new row".
- Even with a hash, two rows are permitted deliberately when the bytes differ.

So two attachments can share a name, and one of them can be several successive
revisions someone uploaded under a single name. Deleting or renaming at the
source would lose content, change `file_id`, and recur on the next hash-less
upload.

There is a second axis. Native paths are one authority namespace across Document
and File Resources — `find_live_path` accepts a `surface` and does
`del surface`, and the live-path unique index carries no surface column — so an
attachment can also collide with a Document at the same path.

## The fix reuses a rule this repo already has

`DocumentService._resolve_free_path` keeps the clean path when it is free and
otherwise steps onto progressively longer slices of the resource's own uuid
(8 -> 12 -> 16 -> full). This wires the same rule into the File surface. The
disambiguator was already being handed in as `resource_id`, so nothing new is
invented and no byte is discarded.

Only the suffix handling differs: Documents hardcode `.md`, while attachments
carry `.txt`, `.html`, `.pdf` or no suffix at all, so the slice is inserted ahead
of the final suffix. For `.md` the result is byte-identical to what Documents
already produce.

Apply migrates vault items before files, so the ordering is deterministic:
Documents keep the path their owner chose, attachments yield.

## Why the resolved path is persisted before publishing

The publication fingerprint includes the path. A retry that re-derived the path
against live state after a partial apply could land on a different one and be
refused as `Native idempotency key was reused with different input`. Writing the
claim first makes a retry replay the same input; the first claim wins, and a
second claim with a different path is an integrity error rather than an
overwrite. `_verify_files` compares against the published path for the same
reason.

## Verification

- New regression test fails on the pre-fix code with the same conflict error and
  passes after.
- Three mutations each fail it: neutering the resolver, reverting verify to
  `logical_path`, and dropping the path persistence.
- Cutover suites: 32 passed, 1 skipped.
- `mypy --python-version 3.14 app/ mcp_server/`: no issues in 365 files.
- `ruff check` clean on the changed files.
- Every other backend suite failure is present, identically, on a pristine
  `origin/main` checkout of the same commit.

## Migration

`100_native_revision_cutover_file_applied_path.py` adds a nullable
`applied_path` column and is registered in `_apply_migrations()`. Existing rows
keep `NULL`, which reads as "published at `logical_path`", so an in-flight
cutover planned before this change verifies unchanged.

### DIFF
--- a/backend/app/db/migrations/100_native_revision_cutover_file_applied_path.py
+++ b/backend/app/db/migrations/100_native_revision_cutover_file_applied_path.py
@@ -1,0 +1,52 @@
+"""Migration 100: record the native path a cutover File was actually published at.
+
+Legacy File names are not unique — the storage key is what
+`UNIQUE(vault_id, s3_key)` covers — and native paths are one authority namespace
+across Document and File Resources. So the path a File is published at can differ
+from the `logical_path` the plan froze, and both `apply` and `verify` need to
+agree on which one it was.
+
+The column is also what makes a retry safe: the publication fingerprint includes
+the path, so re-deriving it against live state after a partial apply would raise
+`Native idempotency key was reused with different input`. Writing the resolved
+path before publishing means a retry replays the same input.
+"""
+
+from __future__ import annotations
+
+import logging
+
+logger = logging.getLogger("akb.migration.100")
+
+
+async def migrate(conn=None):
+    if conn is None:
+        from app.db.postgres import get_pool
+
+        pool = await get_pool()
+        async with pool.acquire() as new_conn:
+            await _run(new_conn)
+    else:
+        await _run(conn)
+
+
+async def _run(conn):
+    if await conn.fetchval("SELECT to_regclass('public.native_revision_cutover_files')") is None:
+        logger.info("Migration 100 skipped: cutover File inventory is absent")
+        return
+
+    async with conn.transaction():
+        await conn.execute(
+            """
+            ALTER TABLE native_revision_cutover_files
+                ADD COLUMN IF NOT EXISTS applied_path TEXT;
+
+            ALTER TABLE native_revision_cutover_files
+                DROP CONSTRAINT IF EXISTS native_revision_cutover_files_applied_path_shape;
+            ALTER TABLE native_revision_cutover_files
+                ADD CONSTRAINT native_revision_cutover_files_applied_path_shape
+                CHECK (applied_path IS NULL OR btrim(applied_path) <> '');
+            """
+        )
+
+    logger.info("Migration 100 recorded the published native path per cutover File")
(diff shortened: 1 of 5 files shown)


imjhyuk-oss:
### The failing check is `main`'s, not this branch's

`Static analysis (ruff + tsc + vitest + secrets)` fails here on

```
FAIL src/lib/__tests__/document-edit-draft.test.ts
  > isolates drafts by user, vault, document, and tab without overwriting another tab
AssertionError: expected { status: 'expired', … } to match object { status: 'restored', … }
  Test Files  1 failed | 115 passed (116)
```

That is not caused by this branch. Evidence:

- `main` has been failing the same check since 2026-09-10T04:17Z; the run before
  it, on 2026-09-09, was green.
- Reverting every file in this branch and re-running the test on a pristine
  checkout of the base commit reproduces it identically (`1 failed | 7 passed`).
- Every file this branch touches is under `backend/`.
- The other four checks pass: backend unit tests, pgvector e2e, frontend e2e.

**Cause.** #515 introduced, in the same commit, both a clamp that shortens a
draft's lifetime to its earliest asset expiry

```ts
for (const value of Object.values(assetExpiresAt)) {
  if (Date.parse(value) < expiresAt.getTime()) expiresAt.setTime(Date.parse(value));
}
```

and a test fixture that pins that asset expiry to an absolute instant:

```ts
assetExpiresAt: { "asset-1": "2026-09-10T01:00:00.000Z" },
```

Once that instant passed, every saved draft is born expired and the assertion
flips. The product logic is right; the fixture is a date bomb.

**It already has an owner, so this branch does not touch it.** The fix exists as
`test(frontend): pin draft isolation clock` on the `feat/akb-208-mcp-catalog-benchmark`
branch (#514), which is why that branch's checks are green. Duplicating it here
would be a second fix for one defect. Worth noting that #514 is still a draft
about something else, so the repo-wide red outlives it until that lands or the
commit is split out.

PR #531, opened 2026-09-11
feat(sso): provider-authoritative email domains for browser login (#529)

### LINKED ISSUE #529: Let a provider be declared the authority for an email domain
## Let a provider be declared the authority for an email domain

An installation that authenticates against a single corporate directory has no
way to say so, and pays for that with a refusal every one of its users hits
once.

### The shape of the problem

With `keycloak_enrollment_mode: invite_only`, a person who already has an
account here signs in through the company IdP, is refused with
`membership_required`, has their arrival recorded, and waits for an
administrator to bind them. Measured on one installation: 76 accounts on the
directory's domain, 68 of them unbound — 68 refusals, each followed by an
administrator approving a match.

The approval applies no judgment. The binding is decided mechanically: the
issuer matches, the email domain is allowed, exactly one active human account
carries that address, and it has no binding yet. We automated exactly that and
the administrator confirms the output. **A step with no judgment is a delay,
not a control.**

Switching to `open` does not help; it makes it worse. The insert collides on
the email unique constraint and the person is refused with
`ExternalIdentityConflictError` instead — locked out rather than linked.

### What the current refusal is defending

The code states it plainly: *"Canonical SSO never adopts an existing account
from mutable profile claims."* The threat is an IdP asserting an address that
belongs to someone else's account.

That threat is real where issuers are many, or where users set their own email,
or where the organisation does not own the mailbox. It is not real where a
single corporate directory is the authority for its own domain: the only party
who could assert someone else's address is the directory administrator, who can
already reset that person's password and sign in as them. The adoption path
grants nothing to an attacker who already holds the IdP.

So the refusal is correct as a default and wrong as an absolute. What is
missing is a way to say "this provider is the authority for these domains".

### Why this is not a fourth enrollment mode

`keycloak_enrollment_mode` is install-wide. Every verified Keycloak principal —
brokered or local-realm — goes through one provisioning path, with no
per-provider branch. So turning on `open` opens the local-realm door too, and
that door must keep behaving exactly as it does. It is not a second identity
provider competing with the corporate one, and it is not the fallback for a
non-SSO installation either — that is a separate auth mode which does not
involve the realm at all. It is the door an `sso` installation keeps to its own
realm, and it exists because an `sso` installation with no external identity
provider would otherwise have no way in. It is therefore where the accounts that
must not depend on the upstream directory live, the recovery admin among them.

Auto-provisioning on that door would contradict its purpose exactly.

One switch cannot describe two doors. The decision belongs on the provider.

### Proposal — one field

```
provider.authoritative_email_domains: list[str]    # default []
```

Evaluated only when no exact `(issuer, subject)` binding exists:

| condition | result |
| --- | --- |
| domain declared, an account has that email | **adopt** — write the binding, sign in |
| domain declared, no account has that email | **provision** — create, sign in |
| domain not declared | unchanged — record the arrival, refuse |

Empty by default, so every existing installation behaves exactly as it does
now, and a provider added later starts closed.

This is enforceable per-provider because the identity token already carries
which upstream the person came through: `identity_provider` is required of a
brokered token, forbidden on a local-realm one, and the login is rejected
unless it matches the provider the flow selected. A second upstream brokered by
the same realm therefore cannot inherit the first one's trust.

### Guards the adoption path still needs

Account ambiguity is impossible — the email column is unique — so four remain:

1. `email_verified` is true (the existing setting already requires it).
2. The target is an active human account. A suspended account must not be
   revived by signing in.
3. The target has no binding for this issuer yet. Note the schema does not
   enforce this: `external_identities` is unique on `(issuer, subject)`, not on
   `(issuer, user_id)`, so it has to be a code check.
4. The target is not the recovery admin. A break-glass path should never be
   claimable by an IdP assertion.

### Notice, not proof

No verification mail belongs here. The address was verified by the directory
that owns the mailbox, the broker is configured to trust exactly that, and the
profile is re-synced on every login. Mailing a corporate address to confirm a
corporate identity is the same empty ritual as the approval above — and this
codebase has no mail path at all: no SMTP settings, no sender.

What is missing is notice. Provisioning already emits an event; adoption should
emit its own, distinct one, so a subscriber can tell "a new person was created"
from "an existing account was claimed". Adoption of an account that holds admin
deserves to be legible in that event — on the installation measured, two of the
76 are admins, and the first arrival the bridge bound was one of them.

Filed from a production cutover where 68 people will otherwise each be refused
once for nothing.

imjhyuk-oss:
Closes #529.

Existing workspace accounts can be locked out after an SSO issuer cutover: `invite_only` requires an exact broker binding, while open enrollment collides with an existing email. This adds opt-in `keycloak_authoritative_email_domains_by_provider` configuration for browser login. A verified login through a configured provider/domain links the existing active human account while preserving its user ID, or provisions an account when none exists. The default remains empty; merge alone does not enable automatic admission.

The callback uses the provider alias selected by the browser flow and verified against both signed tokens. REST/MCP bearer and PAT paths do not gain domain authority. Disabled enrollment remains disabled. Recovery admins, suspended/service accounts, existing issuer bindings, and ambiguous case-insensitive email matches are refused. Email verification follows `keycloak_require_verified_email`; keep it enabled for authoritative domains.

Legacy mixed-case emails resolve to the original account instead of provisioning duplicates. Browser adoption and exact administrative/periodic approval share the subject advisory lock; authority logins additionally acquire an address lock. Same-subject callbacks succeed with one binding, while different subjects cannot automatically claim one account. Successful binding clears pending admission; refused mutations roll back while the pending record commits. Adoption emits `auth.user_adopted`; existing PATs and user-ID-owned grants remain intact.

## Verification

- Latest local focused run: 110 passing tests across authority unit/PostgreSQL, exact identity, and pending admission suites.
- Real PostgreSQL regressions cover simultaneous same-subject callbacks, overlap with periodic approval, mixed-case legacy account continuity, ambiguous case variants, pending cleanup/refusal persistence, and PAT continuity.
- Replaced the deadlocking two-party test barrier with observed PostgreSQL blocking and bounded timeouts. Added the authority PostgreSQL suite to the live DB CI job.
- Current head `c9f9b1b1`: all five GitHub checks pass — backend unit, static analysis, frontend E2E, repository runtime E2E, and all 508 live DB tests.
- Changed-file Ruff and diff checks pass; Helm configuration renders with and without authority domains; design metadata and relative links validated.

## Deployment

Deploy the reviewed backend image and configure the intended provider's corporate domains, for example `entra: [dnotitia.com]`. Preserve `sso`, `invite_only`, and verified-email enforcement. Canary an existing user and verify the same user ID, existing PAT/Vault access, binding, and cleared pending record. Accounts outside declared domains remain subject to explicit admission. This PR does not perform a bulk migration or deploy production configuration.

Design: `docs/design/proposal/2026-09-12-provider-authoritative-email-domains/README.md`.

### DIFF
--- a/backend/app/api/routes/auth.py
+++ b/backend/app/api/routes/auth.py
@@ -544,7 +544,11 @@ async def keycloak_callback(
         expected_provider_alias=provider_alias,
     )
     _require_signed_provider(id_claims, provider_alias)
-    outcome = await project_verified_principal_with_reason(principal)
+    # The alias below is the provider this flow selected (transient) and both
+    # tokens vouched for (biconditional claim check above) — never a value
+    # taken from one token alone. It is the only key the
+    # authority-domain decision reads.
+    outcome = await project_verified_principal_with_reason(principal, provider_alias=provider_alias)
     user = outcome.user
     if user is None:
         # The one place the reason is repeated, and only for the refusal the
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -55,6 +55,10 @@ NATIVE_REVISION_M1_MEASUREMENT_DATABASE_NAME = "akb_revision_m1_measurement"
 
 _DNS1123_LABEL_RE = re.compile(r"^[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$")
 _NATIVE_RUNTIME_IMAGE_DIGEST_RE = re.compile(r"^sha256:[0-9a-f]{64}$")
+# One DNS label of an authoritative email domain: lowercase ASCII
+# alphanumerics with interior hyphens only. Checked after IDNA encoding,
+# so this never sees non-ASCII input.
+_AUTHORITY_DOMAIN_LABEL_RE = re.compile(r"^[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?$")
 
 AuthMode = Literal["local", "sso"]
 LegacyAuthModeStatus = Literal["local_only", "strict_sso", "ambiguous_hybrid", "invalid"]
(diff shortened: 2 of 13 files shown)


imjhyuk-oss:
## Review feedback addressed (b1f0bc0)

Three read-only review lenses (security, correctness, design consistency).
Accepted and fixed:

- **Advisory lock was a no-op**: `pg_advisory_xact_lock` was acquired outside
  any transaction, so it released at statement end. The unbound-identity path
  now runs inside one transaction holding the lock (address-keyed on the
  authority path, subject-keyed elsewhere).
- **Pending admissions never cleared**: adopt/provision success now deletes
  the arrival note, like every other exact-binding writer; a failed adopt
  records the arrival before refusing (audit parity).
- **Adopt race hit 25P02**: the INSERT ran in an already-aborted transaction.
  It now runs under a savepoint so a lost race returns the winner's binding.
  The provision collision handler checks the same-subject fast path first,
  then the authority adopt.
- **IDN declarations were dead config**: config IDNA-encodes but the matcher
  didn't, so a punycode declaration never matched. Both sides now encode
  before comparison; the 253-octet limit is enforced post-encoding; explicit
  null map normalizes to `{}`.
- **Missing coverage**: pending cleanup, failed-adopt arrival, admin
  legibility in `auth.user_adopted` (now also carries `subject`), PAT
  regression across adoption, collision-after-insert adopt, IDN match,
  validator edges.
- **Docs**: `require_verified_email=false` interaction stated on the setting;
  design README updated (concurrency model, failure audit, IDNA both sides).

Deferred (documented, not in scope):

- `verify_access_token` (REST/MCP bearer) still doesn't inspect
  `identity_provider` — the authority decision stays browser-only by
  construction (`provider_alias=None` on all bearer paths). Extending trust
  there is a separate decision.
- Single-`@` assumption in the matcher: a multi-`@` address matches nothing
  (fail-closed); full RFC email validation is out of scope.
- Failed `open`-mode provision on a declared domain (e.g. username taken by
  someone else) refuses with `identity_conflict` rather than adopting — same
  as the existing open path; adopting there would reintroduce the ambiguity
  the guards exist to prevent.
- `values.schema.json` doesn't enumerate most `sso.*` keys (pre-existing
  gap, not introduced here); the new key renders verified with and without
  the map.

PR #532, opened 2026-09-11
fix(search): apply archive scope at hydration, not candidate enumeration (akb#530)

### LINKED ISSUE #530: Default archive scope disqualifies the vault path, so the fast path is never taken
## The default archive scope disqualifies the vault path, so the fast path is never taken

With the readiness fix in place, the vault-granularity search path works — but
only for callers who explicitly ask for `archive_scope=all`. The default shape
of a search still falls back to enumerating candidate source ids, and on a
scope larger than the candidate ceiling that fallback refuses with the
bounded-corpus error.

Measured on one installation, same query, same vault, only the scope differs:

| request | result |
| --- | --- |
| `?q=…` (default) | **422** bounded-corpus |
| `?q=…&archive_scope=all` | **200**, `total_matches=30 returned=5` |
| `?q=…&vault=<50k-doc vault>` (default) | **422** |
| same + `&archive_scope=all` | **200**, `total_matches=9 returned=5` |

### Where it comes from

```python
vault_path_wanted = not (doc_types or source_type or scope != "all") and vault_path_eligible(...)
```

Archive scope defaults to `unarchived`, never `all`, so `scope != "all"` is true
for every ordinary request and the vault path is disqualified before readiness
is even consulted. The fast path exists and nothing reaches it.

The grouping is internally consistent — the archived filter is a document-level
narrowing filter, in the same family as `collection` / `doc_type` / `tags`, and
those genuinely need per-resource ids because the vector store cannot express
them. The vector point carries `chunk_id, source_type, source_id, section_path,
content, chunk_index, dense, indexed_at, vault_id` and no document status.

What makes this one different is that it is the **default**. The others are
opt-in narrowings a caller chose; this one applies to every request nobody
customised.

### What it costs

On the installation measured: 154 archived documents out of 136,183 — **0.11%**.
Excluding that 0.11% sends the other 99.89% down the slow path, and on any scope
over the candidate ceiling it converts a working search into a refusal.

### Suggested direction

The same shape as the `source_type` predicate that already landed: carry the
document's archived state on the vector point and let the driver filter it, so
`unarchived` stops being a reason to enumerate. The plumbing is already there —
`vault_id` and `source_type` both travel that way, every driver implements the
vault filter, and the backfill/readiness machinery for adding a field to
existing points exists and was exercised recently.

If carrying another field is unwanted, the alternative is to stop treating the
default as a narrowing at all: let the vault path run and drop archived hits
during hydration, accepting a slightly short page. That is cheaper to build and
worse to reason about, so the predicate seems the better trade — but either
beats a fast path that the default request can never reach.

Found immediately after the cross-process readiness fix removed the blocker in
front of it.

imjhyuk-oss:
Fixes #530.

## The defect

`archive_scope` defaults to `unarchived`. The vault-path gate demanded
`scope == "all"`:

```python
vault_path_wanted = not (doc_types or source_type or scope != "all") and vault_path_eligible(...)
```

So the fast path had no reachable caller. Every request that nobody customised
fell back to enumerating candidate source ids, and on a scope above the
candidate ceiling that fallback refuses with the bounded-corpus error. The
issue measured it as `?q=…` → 422 against `?q=…&archive_scope=all` → 200, with
archived documents 154 of 136,183 (0.11%).

The grouping was internally consistent — the archived filter really is a
document-level narrowing, in the same family as `collection` / `doc_type` /
`tags`. What makes it different is that it is the **default**. The others are
opt-in narrowings a caller chose; this one applies to every request.

## What changed

The archived predicate moves from candidate enumeration to hydration.

`unarchived` and `all` now take the vault path. `archived` deliberately does
not: it selects **for** the rare tail, so a vault-path top-K would be almost
entirely non-matching and hydration would filter the page down to nothing.
Enumerating ids is the right tool for narrowing to a rare set; it is the wrong
one for excluding it. That split is `archive_scope_allows_vault_path`.

Hydration was already computing the authoritative status for every hit it
returns — `documents.status` on the legacy arm, and on the native arm the
`status` parsed out of the **verified Head body** it fetches anyway. The filter
is one `status_matches` call against a value that was already in hand; it adds
no query and no parse.

A hit excluded by scope must not cost a result slot, so the page is refilled
from the rest of the deduped prefetch pool instead of being returned short. At
the measured density the refill loop essentially never runs, so the common case
still pays exactly one hydration round trip. When the pool is genuinely
exhausted the page is short and says why: the drop is counted as
`archive_scope_excluded` inside the existing `hydration_dropped` degradation
reason.

Applying the predicate at hydration also closes the window where a document is
archived between candidate selection and hydration. On the id path the same
predicate already ran during selection, so the call is idempotent there.

## Why not the other two designs

**Carry archived state on the vector point** (the issue's suggested direction,
mirroring `source_type`) is the symmetric answer and stays the right one if
archived density ever grows. It was rejected on cost, not on shape: it needs a
payload change across all five drivers, a backfill of the existing points, and
a readiness gate — and a readiness gate reintroduces the current symptom for
the duration of the backfill, on an installation that already has a large
derived-index backfill running. Spending that on a 0.11% filter is
disproportionate. Nothing here forecloses it.

**Exclude the archived ids instead of including the candidate ids** looks like
the cheapest exact fix — 154 ids rather than a whole corpus — and it is what I
expected to land. It does not survive contact with the active arm, and this is
the one place the issue is wrong:

> the archived set is 154 rows

That number comes from `documents.status`, and on the native arm that column is
**not the authority**. The retained legacy `documents` rows are a frozen
cutover projection: `native_document_service` writes archived state into the
Head frontmatter and never writes back to `documents`. Measured on the same
installation the issue measured:

| | count |
| --- | --- |
| live native documents | 136,183 |
| `documents.status = 'archived'` | 154 |
| native frontmatter `status: archived` | 151 |
| marked archived in `documents`, `status: draft` in frontmatter | **3** |

Those 3 carry a complete native frontmatter, so the native arm serves them as
`draft`. Excluding by `documents.status` would wrongly hide three live
documents from every search, and the divergence grows with every
post-cutover archive. The authoritative archived set on the native arm exists
only inside the payload bytes, and enumerating it means exactly the envelope
scan the bounded-corpus guard exists to refuse — so the bound would not move to
"archived count" as the issue assumes. It stays at corpus size.

The issue also calls hydration-time filtering "worse to reason about" because
`returned < limit` would mean two things. That objection predates the machinery
that answers it: #520 added per-cause hydration drop accounting and surfaces it
as `degradation_reason`, so a short page now names its cause. With the refill
above, a short page additionally requires an exhausted pool.

## Tests

`backend/tests/test_search_archive_scope_vault_path_unit.py` (new) covers the
hydration filter against the authoritative status, the default that leaves
existing callers unfiltered, the refill, the short page that names its cause,
and the table/file rows that carry no status.

`backend/tests/test_search_filter_contract_unit.py` covers the path selection:
the default request, explicit `unarchived`, and `all` all reach the vault path
and enumerate nothing, while `archived` still enumerates.

Both new behaviours were confirmed by mutation. Against the pre-fix gate the
default and `unarchived` cases fail while `all` passes — the issue's measured
asymmetry, reproduced as a test. With the refill loop disabled, the refill test
returns 2 of 3 requested results.

## Verification

Run locally against this branch and, as a negative control, against a pristine
checkout of the base commit `c25228aa`.

**Backend pytest** (`pytest tests/ -k 'not _e2e'`):

| | failed | passed | skipped |
| --- | --- | --- | --- |
| this branch | 57 | 2798 | 688 |
| pristine `c25228aa` | 56 | 2788 | 688 |

+10 passed is the 11 tests added here minus the one flake below. Diffing the
two failure sets, the 56 shared failures are identical and all live in
`test_git_service.py` (39), `test_external_git_runner.py` (14) and
`test_external_git_capability.py` (3) — none in search. The single extra
failure on this branch is
`test_knowledge_io_unit.py::TestZipRoundTrip::test_zip_is_deterministic`, which
passes in isolation and is a pre-existing flake unrelated to this change:
`bundle_to_zip` calls `ZipFile.writestr` with a `str` arcname, so each entry
embeds `time.localtime()` and two archives are byte-identical only within the
same second. Straddling a second boundary reproduces it on the pristine base
checkout, with none of these changes applied.

**`scripts/check.sh`**: ruff, mypy, bandit, eslint, tsc, the `@akb/client` and
`@akb/markdown-editor` builds/typechecks/vitest, the generated-type drift
check, the packed-SDK consumer proof and the MCP Inspector contract all pass.
It stops at frontend vitest on 1 failing test of 818,
`document-edit-draft.test.ts`, which is also pre-existing and date-dependent:
its fixture hardcodes `assetExpiresAt: "2026-09-10T01:00:00.000Z"`, introduced
by `e86537fc` and now in the past, so the draft loads as `expired` rather than
`restored`. The same single test fails identically on a pristine checkout of
the base commit. No frontend file is touched here.

### DIFF
--- a/backend/CHANGELOG.md
+++ b/backend/CHANGELOG.md
@@ -7,6 +7,35 @@ specifically; the proxy has its own log in
 
 ## Unreleased
 
+### Default archive scope no longer disqualifies the vault path (akb#530)
+
+`archive_scope` defaults to `unarchived`, and the vault-path gate demanded
+`scope == "all"` — so the fast path had no reachable caller. Every ordinary
+search fell back to enumerating candidate source ids, which refuses with the
+bounded-corpus error on any scope above the candidate ceiling. Measured on one
+installation: `?q=…` returned 422 while the identical request with
+`&archive_scope=all` returned 200, with archived documents 154 of 136,183
+(0.11%).
+
+The archived predicate moves from candidate enumeration to hydration, where
+the authoritative status is already parsed: verified Head frontmatter on the
+native arm, `documents.status` on the legacy one. `unarchived` and `all` now
+take the vault path; `archived` deliberately keeps the id path, because it
+selects FOR the rare tail and a vault-path top-K would filter down to nothing.
+
+A hit excluded by scope no longer costs a result slot: the page is refilled
+from the rest of the deduped prefetch pool, and the drop is counted as
+`archive_scope_excluded` in the `hydration_dropped` degradation reason, so a
+genuinely short page (an exhausted pool) names its cause. Applying the
+predicate at hydration also closes the window where a document is archived
+between candidate selection and hydration.
+
+Note for the native arm: the retained legacy `documents` rows are a frozen
+cutover projection and are NOT the archived authority. Measured on the same
+installation, 3 of the 154 rows marked archived there carry `status: draft` in
+their native frontmatter, so filtering on that column would wrongly hide live
+documents.
+
 ### Vault-filter readiness visible to the serving tier (akb#526)
 
 `vault_backfill.is_ready()` was a process-local latch flipped only by the
--- a/backend/app/services/search_service.py
+++ b/backend/app/services/search_service.py
@@ -20,7 +20,7 @@ from typing import Literal
 from app.config import settings
 from app.db.postgres import get_pool
 from app.exceptions import ValidationError
-from app.services.search_filters import ArchiveScope, collection_predicate, escape_like, metadata_matches, resolve_archive_scope
+from app.services.search_filters import ArchiveScope, collection_predicate, escape_like, metadata_matches, resolve_archive_scope, status_matches
 from app.models.document import SearchResponse, SearchResult
 from app.repositories.vault_files_repo import confirmed_file_predicate
 from app.services import sparse_encoder
@@ -394,6 +394,24 @@ def vault_path_eligible(
         and supports_vault_filter(get_vector_store())
         and not (collection or doc_type or tags or source_uris)
     )
+
+
+def archive_scope_allows_vault_path(scope: ArchiveScope) -> bool:
+    """Whether `scope` can be served by the VAULT path, with the archived
+    predicate applied at hydration instead of by candidate enumeration.
+
+    `unarchived` (the DEFAULT) and `all` can. Requiring `all` here is what
+    disqualified every ordinary search from the fast path (akb#530): archive
+    scope is unlike `collection` / `doc_type` / `tags`, which a caller opts
+    into — it is set on every request nobody customised, so treating it as a
+    narrowing filter meant the fast path had no reachable caller at all.
+
+    `archived` cannot, and stays on the id path. It selects FOR a set that was
+    0.11% of the measured corpus, so a vault-path top-K would be almost
+    entirely non-matching and hydration would filter the page down to nothing.
+    Enumerating ids is the right tool for narrowing to a rare set; it is the
+    wrong one for excluding it."""
+    return scope != "archived"
 
 
 def clamp_search_limit(limit: int) -> int:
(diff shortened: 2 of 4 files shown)


imjhyuk-oss:
Merging with four of five checks green.

`backend unit tests`, `pgvector e2e (live DB)`, `shell e2e` and `frontend e2e (mock)` all pass — the first two are the ones this change actually exercises.

`Static analysis` fails on `document-edit-draft.test.ts`, where a fixture hardcodes `assetExpiresAt: "2026-09-10T01:00:00.000Z"` and the calendar walked past it. The same failure is on `main` — its two most recent `check` runs are red the same way while `backend-pytest` and `e2e` pass — and this branch touches no frontend file. Filed as #533 and being fixed separately.

Issue #527, opened 2026-09-11
A NUL byte in a document body abandons its indexing intent silently


imjhyuk-oss:
## A document body containing a NUL byte fails derived indexing forever, and the failure is only a retry counter

During a native-backend index backfill, three documents failed with
`CharacterNotInRepertoireError`. Their bodies contain a `0x00` byte, which a
PostgreSQL `text` column cannot store — confirmed by the server itself refusing
to decode the same bytes:

```
ERROR: invalid byte sequence for encoding "UTF8": 0x00
```

The bodies are ordinary Markdown otherwise; the NUL sits partway through
(offsets 1517, 1778 and 7998 in bodies of 5.6 KB, 5.9 KB and 24 KB). They were
ingested long before the cutover, so the byte survived whatever wrote them.

### What happens now

The derived worker records the exception on the intent, increments
`retry_count`, and schedules another attempt. Every attempt fails the same way,
because nothing about the document changes between them. When the retry ceiling
is reached the intent is abandoned, and the document is then permanently absent
from the derived index: never chunked, never embedded, never returned by ranked
search. It stays readable and greppable, so nothing else signals the loss.

The only trace is `last_error` on one row of an internal queue table. There is
no counter, no log line an operator would look for, and no surface that says
"these N documents could not be indexed".

### Two separable problems

**The byte should not survive ingestion.** A NUL in a Markdown body is not
meaningful content; sanitising it at write time — or refusing the write with a
message naming the offset — costs nothing and prevents the whole class. The
write path already normalises other things about a body.

**A permanently-failing intent should be visible.** A retry counter that climbs
to a ceiling is a fine mechanism, but the terminal state deserves more than a
row nobody reads. An abandoned-intent count exposed alongside the other
indexing stats would let an operator notice at all; naming the resource would
let them fix it.

Of the two, the second matters more: this class of failure will recur for
reasons other than NUL bytes, and the current design loses documents quietly
each time.

imjhyuk-oss:
The three documents have been repaired. **The issue stays open** — the data is
recovered, the two defects are not.

### What the NUL actually was

It stands where a space belongs. Decoding the bytes around one of them:

```
•  대부분의 \0 인공지능 \0 응용분야에서 \0
```

Every NUL sits between words, so the extractor emitted `0x00` instead of `0x20`
while converting a PDF. The prose either side is intact; only the separator is
wrong. That made the repair mechanical rather than a judgement call.

### The repair

[...]

PR #536, opened 2026-09-11
fix(health): report abandoned derived-index intents (akb#527)

imjhyuk-oss:
Addresses the visibility half of #527 — the half that is still broken. The
three documents were already repaired through the API. Deliberately not
`Closes`: the other defect that issue names is untouched here, and there is an
observation about it at the end.

## What was invisible

`NativeDerivedWorker` records an exception on the intent, increments
`retry_count`, and schedules another attempt. At `MAX_RETRIES` it stamps
`delivery_outcome = 'abandoned'` and stops. The Resource is then permanently
absent from the derived index — never chunked, never embedded, never returned
by ranked search — while staying readable and greppable, so nothing else
signals the loss.

Two things kept that from reaching anyone:

- The worker already computed an `abandoned` count, but
  `NativeDerivedWorker.pending_stats` was reachable only from tests and
  `settle()`. Every other queue in this service reports its terminal-failure
  count on `/health` — `vector_store.backfill.upsert.abandoned`,
  `native_file_projection.abandoned`, the delete outbox. The one queue that
  drops whole documents did not.
- `process_once` logs the same `WARNING` for attempt 1 and attempt 8, so the
  moment a Resource is given up on is indistinguishable in a log from the
  transient failures before it.

The consequence is the defect worth naming: `pending` drains to zero whether
the work succeeded or was abandoned, so indexing progress alone reaches 100%
with documents missing from search.

## Counted, beside the progress it contradicts

`native_derived_worker` gains the module-level `pending_stats()` that the other
queue reporters have, returning `pending / retrying / exhausted / abandoned /
applied / superseded / deleted / direct_grep` and a `status`. It is wired into
`/health` as `native_derived` and into `/health/vault/{name}` through
`vault_health()`.

`status` is `degraded` whenever `abandoned` or `exhausted` is non-zero **even
with an empty queue**: a backfill that reached 100% having dropped documents is
not a finished backfill, and the section that reports the progress is the one
that has to say so. This is the shape `native_file_projection` already uses for
the same class of failure, so the two native queues now read alike.

`exhausted` is the separate, pre-terminal state, matching that sibling: the
final claimed attempt while its lease is still in force. The claim query skips
`retry_count >= MAX_RETRIES`, so a process killed exactly there leaves a row
nothing will pick up again until `queue_rescuer` stamps it terminal — a stalled
queue that would otherwise be counted as busy. `retrying` is narrowed to
`0 < retry_count < MAX_RETRIES` for the same reason.

## Named, once, where names already go

The terminal branch of `_failure` now logs an `ERROR` carrying the vault, path,
surface, resource id, revision id, and the exception class. A count tells an
operator that something is wrong; the path tells them what to fix.

It is best effort by construction — the row is stamped terminal before the
lookup runs, so a failed lookup costs a name in a message and never the state —
and it reports the exception **class** only, exactly as `last_error` does,
because an exception message can quote the body that failed to store.

## What this deliberately does not do

- **The intent stays terminal.** #527 raises leaving it claimable instead. For
  this failure class that is worse, not safer: the payload store admits a body
  the derived `chunks.content` column cannot hold, so the intent can never
  succeed and would retry forever. It would also leave `pending` permanently
  non-zero, which destroys the progress signal for every other item in the
  queue. Terminal is the right state; loud is what was missing. Recovery is
  what the issue's own repair did — write a new revision, which supersedes the
  intent and queues a fresh one.
- **No resource names on the world-readable surface.** `/health` is
  unauthenticated, and `/health/vault/{name}` exists because vault existence is
  itself something the access model gates. So the global surface carries counts
  and no identifiers; the per-vault surface narrows the same counts behind the
  reader gate; the identity of the lost Resource lives in the server log, which
  already carries document paths at INFO.
- **The top-level `/health` `status` still reads `ok`.** Making it `degraded`
  would be the loudest possible signal, and it would also change what every
  unauthenticated uptime poller sees. That is an owner's call about an existing
  contract, not a side effect of this fix.

## Tests

Eleven new tests. Ten of them fail against the parent commit; the eleventh is
the guard against over-reporting, and passes there because the parent logs
nothing.

- `test_a_body_the_derived_index_cannot_store_is_lost_but_counted` reproduces
  the real failure against Postgres rather than a stubbed exception: a document
  body carrying one `0x00` where a space belongs, driven through all eight
  attempts. It asserts the durable terminal row
  (`last_error = 'CharacterNotInRepertoireError'`), that no chunk exists for
  the Resource, that `pending` is 0 while `abandoned` is 1 and `status` is
  `degraded`, that the module-level entry point the endpoints call reports the
  same numbers from the same table, and that the log names the path.
- Five unit tests pin the status derivation with no database: abandoned with a
  drained queue is `degraded`, a genuinely clean drain is `ok`, in-flight work
  is `reconciling`, a claim killed on its final attempt is `exhausted` rather
  than `retrying`, and the vault-scoped form narrows by namespace.
- Three unit tests pin the log: the terminal failure names vault and path and
  omits the exception message, a retryable failure stays quiet, and an
  unresolvable Resource still reports the abandonment.
- Two endpoint tests assert `native_derived` reaches `/health` and
  `/health/vault/{name}`.

## Verification

`scripts/check.sh` and the backend suite were both run locally, and both were
already red on the parent commit for unrelated reasons, so the comparison is
against that measured baseline rather than against green.

`scripts/check.sh` — identical on parent and branch: every backend step passes
(ruff clean, mypy `Success: no issues found in 365 source files`, bandit 0 files
skipped) and the run stops at the frontend vitest step with 1 failed
and 817 passed of 818, the single failure being `document-edit-draft.test.ts`,
which is #533 and is being fixed separately. This change touches no frontend
file.

Backend `pytest tests/ -k 'not _e2e'` against a `pgvector/pgvector:pg16`
instance, parent measured on a pristine checkout of `c25228aa`:

| | parent `c25228aa` | this branch |
| --- | --- | --- |
| passed | 3410 | 3421 |
| failed | 60 | 60 |
| errors | 23 | 23 |
| skipped | 39 | 39 |

The failing and erroring sets are byte-identical between the two runs, and all
pre-existing: 39 failures in `test_git_service.py`, 14 in
`test_external_git_runner.py`, 3 in `test_external_git_cascade_unit.py`, 3 in
`test_external_git_capability.py`, 1 in
`test_vault_table_name_collision_unit.py`, and all 23 errors in
`test_access_contributions_postgres.py`. The difference between the runs is
exactly the eleven additional passes.

## One observation about the other half of #527

Not fixed here, and noted because it says why this failure class will recur:
the payload store this write path uses has no NUL check, while
`M1ReferencePayloadStore` rejects `b"\x00"` on both write and verified open,
`native_file_projection.validate_native_text_file` rejects it for text Files,
and the cutover and file-measurement paths reject it too. The `CHECK`
constraint on the payload table is NUL-tolerant by construction, so the byte
has one admitted route in and several closed ones. The PG test added here is
also the evidence for that statement: it gets a NUL body into the store
through the ordinary revision service and watches the derived index fail on
it eight times.

Whether the answer is to reject the write, to sanitise it, or to make the
derived chunk path tolerate a byte the body may legitimately hold, is a
product decision and belongs in #527 rather than in this change.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -471,8 +471,24 @@ async def health(user: AuthenticatedUser | None = Depends(get_optional_user)):
     in one atomic worker), so backfill stats live under
     `vector_store.backfill` and `embed_backfill` is gone — they were
     reporting the same `chunks.vector_indexed_at IS NULL` count.
+
+    Every indexing queue reported here carries its own terminal-failure count
+    beside its progress, and that adjacency is the point: a queue that gave up
+    on an item drains to `pending: 0` exactly like one that finished, so
+    progress alone reads as complete while documents are missing from ranked
+    search. `native_derived` is the document/File-level queue — one intent per
+    Resource revision, and `abandoned` there is a count of Resources that will
+    never be chunked or embedded; `vector_store.backfill.upsert` is the
+    chunk-level one below it; `native_file_projection` is the S3-to-Native
+    admission ahead of both.
     """
-    from app.services import native_file_projection, queue_rescuer, sparse_encoder, vault_backfill
+    from app.services import (
+        native_derived_worker,
+        native_file_projection,
+        queue_rescuer,
+        sparse_encoder,
+        vault_backfill,
+    )
 
     store = get_vector_store()
     vs_info: dict = {"reachable": await store.health()}
@@ -510,6 +526,7 @@ async def health(user: AuthenticatedUser | None = Depends(get_optional_user)):
         "events": await _safe(events_publisher.pending_stats),
         "notifications": await _safe(notification_worker.pending_stats),
         "native_file_projection": await _safe(native_file_projection.pending_stats),
+        "native_derived": await _safe(native_derived_worker.pending_stats),
         "vector_store": vs_info,
     }
 
--- a/backend/app/services/health.py
+++ b/backend/app/services/health.py
@@ -12,18 +12,30 @@ from __future__ import annotations
 
 import uuid
 
-from app.services import embed_worker, metadata_worker, native_file_projection
+from app.services import (
+    embed_worker,
+    metadata_worker,
+    native_derived_worker,
+    native_file_projection,
+)
 
 
 async def vault_health(vault_id: uuid.UUID) -> dict:
     """Return per-vault pending counts. Sequential awaits — the queries
     are sub-millisecond on indexed columns; gather() saves ~1-2ms but
-    adds task-scheduling overhead."""
+    adds task-scheduling overhead.
+
+    `native_derived` is the vault-scoped answer to the question the global
+    counters can only raise: it reports the derived-index intents this vault
+    gave up on, so an operator who sees a non-zero `abandoned` on `/health`
+    can find where the loss is without reading an internal queue table."""
     backfill = await embed_worker.pending_stats(vault_id)
     metadata = await metadata_worker.pending_stats(vault_id)
     projection = await native_file_projection.pending_stats(vault_id)
+    derived = await native_derived_worker.pending_stats(vault_id)
     return {
         "metadata_backfill": metadata,
         "vector_store":      {"backfill": backfill},
         "native_file_projection": projection,
+        "native_derived": derived,
     }
--- a/backend/app/services/native_derived_worker.py
+++ b/backend/app/services/native_derived_worker.py
@@ -30,6 +30,7 @@ from datetime import UTC, datetime, timedelta
 
 import asyncpg
 
+from app.db.postgres import get_pool
 from app.services import delete_worker
 from app.services._backfill import MAX_RETRIES, next_attempt_delay
 from app.services.document_service import _parse_markdown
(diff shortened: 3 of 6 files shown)


imjhyuk-oss:
Merging. `backend unit tests`, `pgvector e2e (live DB)`, `shell e2e` and `frontend e2e (mock)` all pass.

`Static analysis` fails, and the log shows where: `vitest` now reports 818 passed — #535 fixed that — and the run reaches `detect-secrets`, which exits 123 asking for a baseline refresh. That is #537: `.secrets.baseline` pins a false positive in `backend/CHANGELOG.md` by line number, and a changelog grows at the top, so the number goes stale after essentially every backend merge. Nothing in this branch touches it.

Worth noting what that means for the check as a whole: `check.sh` stops at its first violation, so today it has moved two steps — it used to die at `vitest`, now it dies at `detect-secrets`. Each fix reveals the next failure that was queued behind it.

PR #541, opened 2026-09-11
fix(counters): every document counter follows the active authority (akb#525)

imjhyuk-oss:
Completes akb#525. `GET /vaults/{vault}/info` was fixed in #534; it was not
the only counter reading a table the native document path never writes, and
the other two do not read `documents` at all — so a grep for the queries in
the issue does not find them.

## What else was stale, and why grepping missed it

`akb_browse` reports each collection's `doc_count` and `last_updated` from the
denormalised `collections` columns. Those are maintained by
`DocumentRepository.increment_count` / `decrement_count`, which only the
legacy write path calls. On a native installation the browse payload
therefore carried a frozen count directly above the documents contradicting
it — both in one response, which is a sharper contradiction than `/info`,
where the reader cannot see the documents at the same time.

Measured on one installation, with the query this change ships: 25 of 10,344
collections already disagreed with the live ledger — 16 reporting documents
they no longer hold, one reporting zero while holding documents. The count
drifts further on every native write.

The operator corpus inventory (`app/stats`) counted `documents` across the
whole installation, so it reported the pre-cutover total permanently. Same
installation: 136,183 catalog rows against 136,188 live native documents.

## Direction, and why not the other one

The issue names two: the native write path maintains the catalog rows, or the
counters move to the native side. #534 chose the second for `/info`; this
extends that choice to the rest, and the collections finding is what decided
it.

Direction one is not one table. The native path would have to maintain
`documents` *and* `collections.doc_count` / `.last_updated` — the second
denormalised projection the same legacy write path feeds — consistently under
create, update, move, delete and archive. Each is an independent drift
surface, and the browse defect is evidence the drift already happened
silently in a projection nobody thought to check: the issue was found by
reading a zero on one endpoint, and the collection counters had been wrong
the whole time without being noticed.

The native ledger already holds the data in the right shape, with a lifecycle
column, and `search_service`'s hydration already branches on the same
selector these counters now use — so a count can never disagree with what a
search over the same corpus returns.

Leaving the residue unfixed was the worse option specifically because #534
had already moved `/info`: a browse payload whose collection said 0 beside a
vault card that said 2 is a new inconsistency, not a surviving old one.

## Everything that answers "how many" or "how recent" about documents

| surface | reads | state |
| --- | --- | --- |
| `get_vault_info` | `documents` count + latest `updated_at` | branched in #534; moved here into the shared module, query re-shaped (below) |
| `browse` collection items | `collections.doc_count` / `.last_updated` | **fixed here** |
| corpus inventory (`app/stats`) | `COUNT(*) FROM documents` | **fixed here** |
| `/recent` recent changes | `documents` ordered by `updated_at` | already correct — the native revision backend implements `recent_changes` over resource heads |
| `/activity` | git log | already correct — the native backend implements `vault_activity` |
| legacy `DocumentService.browse` | `collections.doc_count` | correct as-is: on a native installation the composition root returns the native document service, so this arm never serves one |
| legacy `create_vault` rollback guard | `COUNT(*) FROM documents` | correct as-is, same reason — the native arm has its own `create_vault` |

All three fixed surfaces now go through one `document_counters` module rather
than three copies of the same branch, so the next counter has one import to
reach instead of its own decision to make. The legacy arm is unchanged on all
three, pays no query at all (the authority is resolved before the pool is
touched), and is pinned by tests.

Two things outside counters, reported separately rather than widened into
this change: `CollectionService.delete` judges emptiness from `documents`,
and the agent-memory reads are legacy-only. Neither is a counter, and both
need their own design.

## Vault last activity: same answer, bounded by the vault

#534's query sorts `native_revisions.occurred_at` across a namespace's live
resources, which makes the planner scan the entire revision ledger — every
vault's history, append-only and never pruned — to answer a question about
one vault. It now orders that vault's own resources and reads the actor off
the head revision.

The two are equivalent because `set_head` writes
`native_resources.updated_at = <the head revision's occurred_at>` in the same
statement that advances `head_revision_id`. Verified on one installation:
136,188 of 136,188 live document resources had a non-null head whose
`occurred_at` equalled `updated_at` and was that resource's newest revision.

For a 50,850-document vault on that installation, `EXPLAIN ANALYZE`:

| form | time | plan |
| --- | --- | --- |
| legacy catalog (pre-#534) | 938 ms | bitmap heap scan over 50,850 wide catalog rows |
| #534 | 93.7 ms | parallel seq scan of all 137,658 revisions + hash join |
| this change | 60.9 ms | index scan of this vault's resources + one head lookup |

Neither native form was a regression against legacy; this one additionally
stops growing with unrelated vaults' history.

The browse recount is a new query where the stored column was free, so it is
scoped the way the render is: browsing into a subtree groups only that
subtree. Unscoped on the same vault it is 118 ms — one grouped index scan,
alongside a collection listing that was already vault-wide.

## The property, and a test that fails without the change

An operator reading a vault's document count and last activity must get
numbers that reflect what the vault contains and when it was last written —
for documents created *and updated* after the cutover, through the ordinary
routes.

The update half is not in the issue, which was written from black-box
measurement of creation. It diverges too: on one installation, documents
repaired through the ordinary update route carried native revisions stamped
that day and catalog `updated_at` values four months old, and the vault
reported the four-month-old date.

`tests/test_document_counters_pg.py` — 9 tests, real Postgres, fresh database
per test, self-skips if unreachable. Writes go through
`NativeRevisionService.create_text` / `replace_text`, the real native write
primitives, not hand-built rows: the last-activity query depends on what
`set_head` maintains, and a fixture that stamped that column itself would be
asserting its own arithmetic.

Against `main` (`f6d39dd9`), 5 fail and 4 pass:

- fail — browse reports the real count and a fresh timestamp; an emptied
  collection reads 0 and not the frozen number; the count stays direct
  children and does not become a subtree total; browsing into a subtree
  counts that subtree (and does not catch a sibling through `_`, which is a
  LIKE wildcard and ordinary in a path); the corpus inventory counts what
  exists (`assert 1 == 3`, the retained catalog row against three live native
  documents).
- pass — the two legacy-arm tests, which must not change, and the two
  vault-info tests. The vault-info pair is a guard for the query re-shape
  above, not evidence of a defect: #534 already satisfies them. Stated
  plainly because a test that passes before the change proves nothing about
  the change.

A cutover-time equality assertion (the issue's suggestion) is deliberately
not added: after the cutover the two totals are *supposed* to diverge, so it
can only be a one-shot check at migration time, not a standing invariant, and
it belongs with the migration rather than with the readers.

## Verification

Run locally, with the baseline established on a pristine checkout of the base
commit (`f6d39dd9`) in a separate worktree.

**`scripts/check.sh`** — green on this branch: E2E suite manifest, agent
roles, ruff, mypy (366 files), bandit, eslint, tsc, `@akb/client` build +
typecheck + vitest + generated-type drift + packed-SDK consumer proof,
`@akb/markdown-editor` build + typecheck + lint + vitest, stdio proxy + MCP
Inspector contract (5/5), frontend vitest (818 passed, 116 files),
detect-secrets. `.secrets.baseline` untouched.

One environment note, since it costs a run to rediscover: with the backend's
runtime dependencies importable, mypy additionally reports a pre-existing
`list` invariance error in the qdrant vector store, on a line this change does
not touch. It reproduces identically on the pristine baseline, and CI cannot
see it by construction — the check workflow installs only the four analyzers,
so `qdrant_client` is missing and `ignore_missing_imports` makes the argument
`Any`. The green run above uses exactly CI's analyzer set at CI's pins.

**Backend suite** (`pytest tests`, real Postgres reachable so the `*_pg.py`
files run rather than skip):

| | passed | failed | errors | skipped |
| --- | --- | --- | --- | --- |
| baseline `f6d39dd9` | 3490 | 63 | 40 | 48 |
| this branch | 3499 | 63 | 40 | 48 |

Identical failure and error counts, and identical per-file breakdown —
`test_git_service.py` 39, `test_external_git_runner.py` 14, the external-git
cascade/capability files 6, `concurrency/test_ensure_collection_race_unit.py`
3, `test_vault_table_name_collision_unit.py` 1; errors are
`test_access_contributions_postgres.py` 23 and `mcp_e2e/*` 17. None is
touched by this change. The +9 passed is exactly the new file.

No schema migration, no config change, no response-shape change.

---

### DIFF
--- a/backend/CHANGELOG.md
+++ b/backend/CHANGELOG.md
@@ -7,6 +7,46 @@ specifically; the proxy has its own log in
 
 ## Unreleased
 
+### Every document counter follows the active authority (akb#525)
+
+`GET /vaults/{vault}/info` learned to read the native ledger on
+`postgres_native`, but it was not the only counter answering from a table the
+native write path never touches, and the other two do not read `documents` at
+all — a grep for the query in the issue does not find them.
+
+`akb_browse` reports each collection's `doc_count` and `last_updated` from the
+denormalised `collections` columns, which only `DocumentRepository.increment_count`
+/ `decrement_count` bump — legacy-write-path-only. On a native installation the
+browse payload therefore carried a frozen count directly above the documents
+contradicting it. Measured on one installation: 25 of 10,344 collections already
+disagreed with the live ledger — 16 reporting documents they no longer hold, one
+reporting zero while holding documents.
+
+The operator corpus inventory (`app/stats`) counted `documents` across the whole
+installation, so it reported the pre-cutover total permanently. Same
+installation: 136,183 catalog rows against 136,188 live native documents, the
+gap widening with every write.
+
+Both now read the authority that holds the documents, alongside the vault-info
+counters, through one `document_counters` module rather than three copies of the
+same branch. The legacy arm is unchanged on all three, and pays no query at all
+— the authority is resolved before the pool is touched. Browsing into a subtree
+scopes the recount to that subtree, since only its collections are rendered
+(unscoped on the 50,850-document vault: 118 ms).
+
+Vault last activity also stops sorting `native_revisions.occurred_at`, which made
+the planner scan the entire append-only revision ledger — every vault's history,
+not the one being asked about. It orders the vault's own resources instead and
+reads the actor off the head revision, which `set_head` keeps in step by writing
+`updated_at` and `head_revision_id` in one statement (verified: 136,188 of
+136,188 live document resources had a head whose `occurred_at` equalled
+`updated_at` and was that resource's newest revision). For a 50,850-document
+vault: 93.7 ms to 60.9 ms, and no longer growing with unrelated vaults' history.
+
+Counting is unchanged in meaning on both arms: archived documents still count
+(native `lifecycle` records deletion, not archival), and a collection's count is
+still its direct children.
+
 ### Default archive scope no longer disqualifies the vault path (akb#530)
 
 `archive_scope` defaults to `unarchived`, and the vault-path gate demanded
--- a/backend/app/services/access_service.py
+++ b/backend/app/services/access_service.py
@@ -35,6 +35,7 @@ from app.services.access_contributions import (
 )
 from app.services.account_markers import is_retired_recovery_admin_password
 from app.services.account_service import presented_issuer_or_none
+from app.services import document_counters
 from app.services.edge_boundary import edge_scope_sql, vault_uri_prefix
 from app.services.role_sync import get_role_sync
 from app.services.uri_service import vault_uri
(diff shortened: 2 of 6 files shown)


imjhyuk-oss:
Merging — all five checks green.

One correction to the PR body for anyone reading it later. It states that 25 of 10,344 collections already carried a wrong `doc_count` on the installation measured. That does not reproduce here. Measured three ways against the live database:

| basis | mismatches |
| --- | --- |
| `collections.doc_count` vs `documents` | 0 |
| `collections.doc_count` vs live native resources | 0 |
| `collections.last_updated` vs newest native revision | **805** |

So the counts are intact and it is the timestamps that have drifted — on 805 collections rather than 25. The direction the PR argues is unaffected and the fix covers both fields; only the number and which field it applies to are different.

Filed separately from the same enumeration: #542 (`agent_memory_service` reads only the legacy catalog) and #543 (`CollectionService.delete` builds the legacy repository with no backend dispatch).

PR #544, opened 2026-09-14
feat(sso): support account completion for registered companion apps

kwoo24-oss:
An existing AKB account can remain unbound when its owner signs in through a companion application. This adds optional server-to-server account completion after a registered companion backend verifies its own browser-bound OIDC login, reusing AKB's authoritative-provider account adoption policy.

Completion is an onboarding/migration capability. Already linked users can sign in through their companion's ordinary OIDC flow and `/auth/me` without this endpoint or its RSA signing key. Operators can enable the capability for a transition period or retain it for ongoing onboarding.

## Changes

- Add `POST /api/v1/auth/sso/companion/complete`, registered by OIDC client ID, provider aliases and dedicated RSA public keys. AKB independently verifies the BFF signature and Keycloak access/ID token pair.
- Reuse `project_verified_principal_with_reason` for account adoption/enrollment, account guards, pending admissions and role synchronization. `auth_service.py` and ordinary bearer/PAT/MCP policy remain unchanged.
- Add migration 101 for consumed assertion JTIs. Responses contain the resolved user and stable account-denial codes, with no AKB cookie/token or completion-result cache.
- Publish required credentials and response models in OpenAPI, preserving sanitized missing/duplicate-header rejection and credential redaction.
- Provide product-neutral configuration, a complete integration guide, and optional Helm `sso.companionClientIdsByOrigin`/`sso.companionLoginClients` maps. Both default to empty; only public registration data is stored on AKB.
- Document activation and retirement. Removing completion registration retains identity bindings; normal SSO still requires the separate human API client allowlist. No new feature flag is needed.
- Include PostgreSQL regressions in the live DB CI job, including fresh bearer/HTTP `/auth/me` login after registration removal, disabled completion refusal and no adoption of unlinked accounts through bearer auth.

This is a custom AKB extension on top of OIDC. The BFF signature delegates browser state/PKCE verification to a registered BFF; AKB independently verifies the user token pair and provider. Browser navigation stays between the application and its identity provider.

## Deployment and integration

To enable completion, deploy AKB and migration 101, then register the reviewed public key/client/providers before enabling the companion callback integration. Preserve authoritative-domain and verified-email policy. Register every exposed provider, including `local` when workspace sign-in is enabled. Exclude credential headers from ingress/APM capture and keep clocks synchronized.

After the intended identities are linked, disable completion in every companion instance and redeploy it first. Verify a new OIDC login through `/auth/me`, drain old callbacks, then remove that client's completion registration. Keep the ordinary API client allowlist and identity data; migration 101 is not rolled back. The private signing key can leave active deployment secrets once unused. Future unlinked accounts need reactivation or approved provisioning/linking; existing open-enrollment/JIT policy is unchanged.

Any reviewed BFF using the configured Keycloak issuer can implement this contract. The account policy builds on #531; [Reef #462](https://github.com/dnotitia/reef/pull/462) is one consumer, not an AKB dependency.

## Validation

- Current lifecycle follow-up: 97 PostgreSQL, PR #531 policy, signature/token, OpenAPI and Helm tests passed against isolated PostgreSQL. Tests verify the original account ID survives completion deregistration and unlinked bearer auth does not adopt an account. Ruff, YAML parsing and `git diff --check` passed.
- Initial implementation: 185 focused PostgreSQL, policy, token/signature, concurrency, pending-admission and redaction regressions passed. The public-contract follow-up passed 64 credential, OpenAPI and Helm tests.
- Earlier full backend Ruff, changed-source mypy and local documentation links passed. Bandit reported zero medium/high findings and zero skipped files. This lifecycle follow-up changes tests and documentation, not backend service code.
- `scripts/check.sh` was rerun and remains blocked by missing dependencies in four untouched Node projects. Earlier full backend mypy also reported one existing error in untouched `app/services/vector_store/qdrant.py:399`; changed sources pass.
- No production deployment or live identity-provider sign-in was performed.


younglokim-oss:
After reconsidering the product policy, we have decided not to adopt this change. AKB will remain the primary application and authority for account onboarding and identity linking, while companion applications will use ordinary OIDC and `/auth/me` only for accounts that are already linked. Delegating account-completion authority to registered BFFs would introduce dedicated RSA keys, replay storage, client/provider registration, and ongoing operational and audit costs that are not justified by the current product requirements. Provider-authoritative adoption, when needed, should remain in AKB’s own browser callback and existing account policy; we will not add a companion server-to-server completion API.

PR #554, opened 2026-09-15
feat: add local and managed SSO account lifecycle

kwoo24-oss:
## Summary

Add a Security section for ending all local login sessions or ordinary AKB SSO browser sessions. Session revocation includes the current session and preserves PATs. Local account deletion requires the current password and account name, blocks owned vaults (including archived vaults), protects recovery identities and the last eligible local administrator, and preserves shared publications. SSO accounts display an organization-managed notice and cannot self-delete.

SSO logout preserves the Keycloak session, so users can intentionally sign in again. A database sequence rejects callbacks from login flows started before logout. The browser pins actions to the reviewed CSRF cookie and ignores stale completion responses without clearing a newer session.

An independently enabled worker reconciles existing Keycloak identities in bounded pages. A confirmed disabled or deleted Keycloak user suspends the AKB account and revokes its PATs and browser sessions while preserving vaults and identity bindings. Management API failures preserve account state. Reactivation remains an explicit administrative action.

Failed pages advance the scan and switch subsequent reads to one identity at a time. Failed identities are retried on the next sweep, without blocking their neighbors or later pages. A complete clean sweep restores batches of 25 and clears the durable health error; successful individual pages and process restarts do not erase unresolved errors.

The server rechecks identity and deletion conditions inside one transaction. An independent outbox retries PostgreSQL role cleanup after deletion. The UI pins the request to its authentication snapshot, verifies success receipts before clearing credentials, and handles timeout, reauthentication and unsupported-server states.

Local JWTs use a monotonic session generation. A database trigger also fences legacy cutoff-only revocation; password confirmation follows the same Unicode normalization as registration and login. The design document, example configuration and changelog describe the public contract and rollout.

## Compatibility and rollout

- `DELETE /api/v1/my/account` now returns **410**. Clients must preview the lifecycle contract and use `POST /api/v1/my/account/deletion`.
- Apply migrations **101–103** before serving the new contract. The cutoff trigger must be installed before new JWT issuers accept requests; replace all SSO callback handlers before enabling SSO self-service.
- Self-service is **disabled by default**. Replace/drain old issuers and verifiers, verify the cleanup worker heartbeat, then set `account_self_service_enabled: true`.
- Keycloak synchronization is separately disabled by default (`sso_account_sync_enabled: false`), requires the direct management adapter with realm/user read permissions, and checks up to 25 identities per tick (30 seconds by default). Detection is eventual over a full sweep. SSO self-deletion remains unsupported.
- This PR does not bump versions or deploy services.

## Validation

- `bash scripts/check.sh`: passed, including frontend 129 files / 978 tests, static analysis, SDK checks and secret scanning.
- Combined SSO lifecycle, managed-account synchronization, browser boundary, local lifecycle and worker regression suite: **79 passed**, including real PostgreSQL, ASGI cookie/CSRF requests and mocked Keycloak management HTTP responses. A live Keycloak deployment was not exercised.
- Follow-up failure-isolation regression suite: **42 passed**, covering persistent per-user errors, page timeouts, neighboring/later account processing, retry recovery, failed final pages and durable health across restarts.
- Real PostgreSQL lifecycle/generation tests: **22 passed**; local JWT unit tests: **28 passed**.
- Real Chromium account lifecycle E2E: passed (two JWTs revoked, PAT retained, reauthentication, cancellation focus, deletion invalidates JWT and PAT).
- Isolated repository runtime gate: **17 MCP pytest cases + 546 curated HTTP assertions passed**, including the JWT HTTP E2E (19 passed).
- Additional hybrid access/ownership lifecycle suite: **17 passed**, including safe fixture account cleanup.
- Independent backend and frontend reviews: no outstanding findings; the generation compatibility and Unicode regressions have tests.


kwoo24-oss (commented):
Reviewed head `9d1470402cdf3832c8ed312c4471f7e6f28a0bd7` against the PR base. The published diff matches the locally reviewed diff; the intervening Markdown editor changes on main do not overlap this change.

**No outstanding actionable code findings.**

Review covered identity-bound mutations, ownership and administrator row-lock races, atomic deletion, durable role cleanup, generation-based revocation and legacy cutoff compatibility, password normalization, auth snapshot handling, unsupported contracts, timeout and cancellation. Independent backend and frontend passes reached the same conclusion. Public documentation and changed-file secret checks are clean.

Validation: repository static/SDK/frontend gate passed; 22 real PostgreSQL regression tests and 28 local JWT unit tests passed; isolated runtime passed 17 MCP cases and 546 HTTP assertions; the additional hybrid access/cleanup suite passed 17 assertions. Real browser lifecycle coverage also passed.

Rollout constraints remain intentional: the old self-delete endpoint returns 410; migrations precede new JWT issuance; old issuers/verifiers must be drained; self-service stays disabled until explicitly enabled with a healthy cleanup worker; SSO is unsupported.

GitHub CI is still running at the time of this review. This comment records code review and local validation, not a merge approval.

kwoo24-oss (commented):
Reviewed the SSO extension at commit 0596e2f1. No actionable findings remain.

Checked the server-bound login sequence, callback/logout serialization, rejection of pre-logout callbacks, CSRF and expected-user checks, transactional audit rollback, and migration 103 parity with fresh initialization. Also checked managed-account synchronization: authoritative realm/user reads, failure behavior, runtime and identity revalidation, suspension with PAT/browser revocation, and vault preservation.

Validation: 79 combined backend regression tests passed, including real PostgreSQL; scripts/check.sh passed with 978 frontend tests. Keycloak management HTTP behavior was tested with mocks; a live Keycloak deployment was not exercised. Both rollout flags remain disabled by default. No deployment was performed.

kwoo24-oss (commented):
## Review of 0596e2f1

One actionable finding; recommend fixing before merge. This supersedes the earlier no-findings assessment for the synchronization worker.

### [P2] Keep the account sweep progressing after a per-user management failure

Location: backend/app/services/sso_account_sync.py, lines 100–104 (and the page-wide read at line 65).

The cursor only advances after the entire page succeeds. read_managed_account_states rejects the whole page when any one user read fails, so a persistent error for one subject causes every tick to retry the same 25 identities. Every later identity is permanently excluded from reconciliation, including users disabled in Keycloak whose AKB PATs should be revoked. A page-wide timeout has the same effect.

Reproduced using the actual process_once function with mocked database/control boundaries: 26 ordered identities, one persistent management error in the first page and a disabled 26th identity. Three ticks each requested the identical 25 subjects, the cursor stayed None, and the disabled subject was never read.

Preserve accounts with uncertain reads, but isolate their retries so the scan continues to other identities; retain degraded health until failed reads recover. Add a regression proving that an error in the first page does not indefinitely block a disabled identity on a later page.

Other review coverage: login sequence/callback fencing, CSRF and identity binding, concurrent browser-session revocation, frontend same-account/new-cookie changes, and late success/401 responses. No additional reproducible findings in these areas. Previously recorded test passes do not cover the failure-isolation scenario above. CI is still pending except for frontend mock E2E, which passed.

kwoo24-oss (commented):
Addressed the P2 synchronization starvation finding in b774cf5b.

Failed pages now advance the cursor and switch the scan to individual reads. The next sweep retries each failed-page identity independently, so one failed subject cannot indefinitely block its neighbors or later pages. A fully clean sweep restores batches of 25. The durable health error is only cleared at a clean sweep boundary, including after a worker restart.

Added real-PostgreSQL regression coverage for persistent management errors and timeouts, a disabled 26th account, a neighbor from the original failed page, credential preservation for the uncertain account, recovery to batch mode, a failed partial final page, and restart-safe health. Related backend suite: 42 passed. scripts/check.sh: passed, including 978 frontend tests. Follow-up review found no outstanding boundary defects; page_size remains the configured maximum for consistent reporting across separate API/worker processes.

No deployment was performed. Live Keycloak integration remains unverified; management HTTP behavior is mocked in the tests.

kwoo24-oss (commented):
Re-reviewed PR head b774cf5b17d57dc0769a41e45ab2d7622f97478a. No new actionable findings or code-review blockers found; the earlier P2 synchronization-starvation finding is addressed.

Checked partial processing failures, status-write failures, full and partial final pages, repeated timeouts, authority changes, clean-sweep-only error clearing, and separate API/worker health reporting. Verified the new regression tests are included in the live-PostgreSQL CI suite.

Validation evidence remains the passing 42-test related backend suite and scripts/check.sh (including 978 frontend tests); unchanged tests were not rerun for this read-only review. Live Keycloak integration remains unverified. CI has not yet completed all checks, so this review does not assert a fully green CI result. No code changes or deployment were made.

Issue #608, opened 2026-09-18
A short page has no field that explains why it is short


imjhyuk-oss:
## The gap

A caller that asks for `limit` results and receives fewer has nothing in the
response that explains the difference.

`total_matches` cannot do it. It is the prefetch pool size measured before
hydration, and the model comment says so in as many words —
`backend/app/models/document.py`: "**It is NOT a corpus-wide hit count.**"
`truncated` answers a different question again: whether the pool hit its ceiling.

So when the default `unarchived` scope removes a candidate at hydration
(`backend/app/services/search_service.py:1665`), the page can come back short and
the response says nothing about why. Not because the information is unavailable —
it is counted, logged, and returned from `_hydrate_hits` as `dropped` — but
because nothing carries it to the caller.

That gap is why the archive cause was wired into `degraded` in the first place.
akb#604 removed it from there, correctly, without putting anything in its place.

## What general-purpose engines put here

A count, not a flag. Elasticsearch reports `hits.total.value` with
`hits.total.relation` (`eq` / `gte`) so the client knows both the size of the
result set and whether that size is exact, and it reports skipped and failed
shards as structured counts under `_shards` rather than as booleans. Solr reports
`numFound` beside `partialResults`.

The shape to copy is a structured count that names its cause, not another
boolean.

## Proposal

Add one field to `SearchResponse` carrying the exclusions that are not faults,
keyed by cause:

```
excluded: {"archived": 1}
```

Empty when nothing was excluded. The sum is the count a caller needs to explain a
short page; the key says which filter did it, which is what distinguishes "your
scope removed it" from "something went wrong".

Four things that should hold:

1. **Only non-fault causes belong here.** `NON_DEGRADING_DROP_CAUSES`
   (`search_service.py:88`) already draws that line — this field is its natural
   counterpart. The five genuine causes stay in `degradation_reason`, so each
   fact lives in exactly one place.
2. **The keys are public vocabulary, not internal ones.** The internal counter
   key is `archive_scope_excluded`; the response should say `archived`. Internal
   diagnostic strings changing shape should not be a breaking API change.
3. **It counts candidates considered for this page, not the whole pool.**
   `dropped` accumulates across the initial hydration and the refill loop, so it
   is already page-relative — that is the number a caller can act on, and the
   field should not claim to be anything wider.
4. **No name from the "total" family.** `total_matches` has already been
   misread once as a corpus count, which is why its own docstring shouts
   otherwise. A new field should not invite the same reading.

[...]

Issue #605, opened 2026-09-18
A degraded search that carries results is discarded by the editor search adapter


imjhyuk-oss:
## The narrow defect

`frontend/src/lib/markdown-adapters.ts:295` throws on any degraded response:

```ts
const response = await searchDocs(query, vault, 20, {}, { signal: context?.signal });
if (response.degraded) throw new Error("Search results are temporarily unavailable.");
return response.results...
```

The throw happens before `response.results` is read.

Its test pins the case the guard was written for, and only that case —
`frontend/src/lib/__tests__/markdown-adapters.test.ts:144-151`,
"does not present degraded retrieval as a genuine empty result", mocking
`{ degraded: true, results: [] }`. For an empty degraded response the throw is
right: an empty list would read as "nothing matches", which is the failure mode
the guard exists to prevent.

For a degraded response that *carries* results it is not right, and nothing
tests that case.

## Why the distinction matters

The backend deliberately preserves results in one degradation mode and says so
at the return site (`backend/app/services/search_service.py:1775-1777`):

```python
# `sparse_reason` is None on the normal path; set when the sparse leg
# was down and we ran dense-only (degraded-but-has-results).
return hits, sparse_reason
```

Under `sparse_encoder_degraded` the keyword leg failed and the search ran on the
dense leg alone. The hits are real and ranked; one retrieval leg is missing, so
the ordering may differ from a healthy search. The other two reasons return `[]`
(`:1751`, `:1782`), so for them nothing is discarded and the guard is a no-op.

So the throw is inert in the cases with nothing to show, and destructive in the
one case where the backend went to the trouble of keeping something — and that
case is a partial outage, when falling back to "no results at all" costs the
most rather than the least.

Narrowing the guard to `response.degraded && response.results.length === 0`
preserves the existing test's intent exactly and closes the case it does not
cover.

## The related question, which is a decision and not a defect

`frontend/src/components/global-search-dialog.tsx:187-189` stores the results and
then makes them unreachable:

```ts
setResults(nextResults);
setActiveIndex(nextResults.length > 0 ? 0 : -1);
if (response.degraded) setError("Search is incomplete; the retrieval service is temporarily unavailable. Please retry.");
```

Unlike the adapter, this is deliberate. Its test says so in its name —
`frontend/src/components/__tests__/global-search-dialog.test.tsx:90-104`,
"does not expose or activate hidden partial results during a degraded response" —
and asserts the combobox collapses (`aria-expanded="false"`, no
[...]

Issue #604, opened 2026-09-17
An archived document filtered as requested reports the retrieval service as unavailable


imjhyuk-oss:
## What happens

A search that excludes an archived document — exactly as the default `unarchived`
scope is supposed to — comes back with `degraded=true` and
`degradation_reason: "hydration_dropped:archive_scope_excluded=1"`.

The page can be complete when this happens. The refill loop pulls a replacement
from the spare pool, so the caller receives a full page of correct results and is
told the retrieval service is unavailable.

Raising `limit` on one unchanged query is enough to flip it:

```
limit small enough that the archived hit stays out of the page
  -> degraded=false
limit large enough that it enters the prefetch pool, with spare left to refill
  -> degraded=true, reason hydration_dropped:archive_scope_excluded=1
     returned == limit
```

Same query, same corpus, same moment. The only thing that changed is whether an
archived document happened to land in the pool — and in the second case nothing
is missing from the response.

## Why this is not cosmetic

Two of the three consumers throw the results away.

- `frontend/src/lib/markdown-adapters.ts:295` —
  `if (response.degraded) throw new Error("Search results are temporarily unavailable.")`,
  before `response.results` is read at all.
- `frontend/src/components/global-search-dialog.tsx:189` — sets an error after
  storing the results; `frontend/src/components/__tests__/global-search-dialog.test.tsx:90-104`
  pins that the combobox collapses and the results become keyboard-unreachable.
- `frontend/src/pages/search.tsx:911-919` — renders the banner. It branches on
  `degradation_reason` for exactly one value, so this reason takes the `else`:
  "Search is incomplete. The retrieval service is temporarily unavailable."
  The result count is replaced by "Search incomplete" (`:482-483`) and screen
  readers are told results may be missing (`:501-502`).

Agents get the same claim from the tool description at
`backend/mcp_server/tools.py:400-403`: "the retrieval index hit a **transient**
failure (vector-store outage or a degraded leg) … **Retry shortly**, or fall back
to akb_grep." Retrying a deterministic scope filter returns the identical response.

## Root cause

`backend/app/services/search_service.py:1132-1136` promotes any non-empty drop set
to the degradation reason:

```python
hydrate_reason = (
    f"hydration_dropped:{...}" if dropped else None
)
if hydrate_reason is not None and degraded_reason is None:
    degraded_reason = hydrate_reason
```

`_hydrate_hits` produces six causes. Five mean something is wrong or stale —
[...]

PR #613, opened 2026-09-18
fix(search): degraded means the response is incomplete

### LINKED ISSUE #611: A fault the refill loop compensated still marks a complete page degraded
## What happens

A fault that the refill loop fully compensated still marks the response
`degraded`. The page is complete — `returned == limit`, nothing missing — and it
reports that the result set is incomplete.

Reproduced live against a deployment carrying akb#604 and akb#608:

```
degraded=True  reason=hydration_dropped:hydration_miss=4
returned=5/5   total_matches=24   excluded={}
```

Eleven consecutive observations during one window, every one of them a full
page. The recipe is below; it needs no outage, only an ordinary delete.

## Why this is the same defect akb#604 fixed, one cause over

akb#604 established that `degraded` must mean a failure, and carved
`archive_scope_excluded` out because a filter is not a failure. That was half the
rule. The other half is in the field's own documented meaning, reconciled by the
same change (`backend/app/models/document.py`):

> `degraded` is true when something FAILED **and the result set is therefore
> incomplete/empty**

Two claims are bundled there, and the second does not follow from the first. The
refill loop at `backend/app/services/search_service.py:1160-1167` exists precisely
to replace dropped candidates from the spare pool, and when it succeeds the
caller loses nothing. But the flag is set from the presence of a fault, never
reconciled against what actually came back:

```python
faults = {k: v for k, v in dropped.items() if k not in NON_DEGRADING_DROP_CAUSES}
hydrate_reason = (
    f"hydration_dropped:{...}" if faults else None
)
```

No length is consulted.

## Why it fires more than an outage would

`hydration_miss` means a hit survived retrieval but its source row was gone by
hydration — a document deleted, or a derived chunk whose head moved, in the
window between the two. That is not an outage. It is the ordinary race between
search and writes, and it opens every time a document is deleted or replaced.

So the dominant trigger for `degraded` is now a write race rather than a
component failure, and in the observed window it produced complete pages 11 times
out of 11.

## What that costs

`degraded` is not decorative. Consumers act on it: one web surface renders a
warning and shows the results, another sets an error that empties the list
(`frontend/src/components/global-search-dialog.tsx:187-189`, where
`visibleResults = loading || error ? [] : results`), and the MCP tool description
tells agents the failure is transient and suggests retrying or falling back to a
literal search. On a complete page all of that is wrong, and retrying a race
that has already resolved returns a response that is no longer degraded — which
teaches callers that the flag means nothing.

## Suggested fix

Report the two facts separately, the way akb#608 separated a filter from a fault.

**`degraded` should mean the response is incomplete.** A fault whose drop the
refill loop replaced did not make it incomplete, so it should not set the flag.
A fault that left the page short did, and should.

**A compensated fault should still be visible**, because "a chunk pointed at a
row that is gone" is a corpus-integrity signal worth keeping even when the page
filled. Dropping it entirely would trade one blind spot for another. The natural
home is a count beside `excluded`, keyed by cause, so each fact keeps exactly one
place:

- `excluded` — the request removed it (a filter); retrying changes nothing
- the new count — a fault removed it but the page recovered; the corpus has a
  stale pointer worth chasing
- `degraded` — the response is genuinely short or empty

The internal `dropped` dict already carries every cause and should not change.

## Reproducing

No outage required. On any deployment:

1. Create enough documents sharing a unique term that the prefetch pool exceeds
   `limit` comfortably.
2. Wait for them to become searchable.
3. Delete about half of them.
4. Search for the term at a small `limit`, repeatedly, for a few seconds.

During the window between the delete and its propagation to the chunk store, the
responses carry `degraded: true` with `hydration_dropped:hydration_miss=N` and
`returned == limit`.

imjhyuk-oss:
Closes #611.

`degraded` now means the response is genuinely incomplete, which is what its own
documentation on `SearchResponse` has always said it meant. A hydration fault
whose dropped candidate the refill loop replaced no longer sets it; a fault that
left the page short of the requested `limit` still does. The fault is not thrown
away with the flag — a complete page reports it as a new `recovered` count.

## Why the flag was wrong

The flag was set from the *presence* of a fault and never reconciled against
what came back:

```python
faults = {k: v for k, v in dropped.items() if k not in NON_DEGRADING_DROP_CAUSES}
hydrate_reason = f"hydration_dropped:{...}" if faults else None
```

No length is consulted. The refill loop directly above it exists precisely to
replace dropped candidates from the spare pool, so a full page and a short one
produced the same flag.

That matters more than an outage would, because the dominant cause is not an
outage. `hydration_miss` means a hit survived retrieval and its source row was
gone by hydration — the ordinary race between a search and a concurrent delete
or replace — and it opens on every such write. The flag's consumers act on it:
one surface renders a warning, another sets an error that empties the result
list, and the MCP tool description tells agents the failure is transient and
suggests retrying. On a complete page all of that is wrong, and retrying a race
that has already resolved returns a response that is no longer degraded, which
teaches callers the flag means nothing.

## The predicate, and why it is the right one

A fault degrades the response iff `returned < limit`, i.e. the page came back
short of what was asked for.

That is not an approximation of "did the fault cost the caller a result"; it is
equivalent to it, and the refill loop is why. The loop drains the *entire* spare
pool before it will return a short page:

```python
page, spare = unique_hits[:limit], unique_hits[limit:]
results, dropped = await self._hydrate_hits(page, archive_scope=scope)
while len(results) < limit and spare:
    ...
```

It exits in exactly one of two states. Either `len(results) == limit` — the
caller received every result they asked for, so the cost is zero whether one
candidate was dropped or twenty. Or `spare` is empty with the page still
unfilled — the whole pool was consumed, so the page holds `total_matches` minus
every drop, where without the faults it would have held
`min(limit, total_matches - filtered)`, which is strictly larger whenever a
fault occurred. So `returned < limit` with a fault present means the fault cost
at least one result, and `returned == limit` means it cost none. There is no
third case.

A naive `returned < limit` on its own would be wrong in the other direction. A
corpus holding fewer matches than `limit` returns a short page with nothing
wrong with it, and a page shortened by the archive filter is #604 all over
again. The conjunction is what names the fault case and only the fault case;
there is a test for each side.

Retrieval-leg reasons are untouched: a leg that raised still degrades and still
outranks the hydration causes in the reason string. Under
`sparse_encoder_degraded` a whole retrieval leg is missing from the ranking,
which is incompleteness of a kind no refill can repair.

## Where the fault goes instead

A chunk pointing at a row that is gone is a corpus-integrity signal worth
keeping even when the page filled, so dropping it with the flag would trade one
blind spot for another. A complete page now reports it as a count beside
`excluded`:

```
recovered: {"hydration_miss": 2}
```

`{}` when the page needed no repair, and always present, so a caller tests it
for emptiness rather than for presence — the same guarantee `excluded` carries.

A drop is now reported in exactly one of three places, chosen by what it did:

- `excluded` — the request removed it (a filter). Retrying changes nothing.
- `recovered` — a fault removed it and the page recovered. Nothing was lost to
  the caller; it is corpus health, not a problem with this response.
- `degraded` / `degradation_reason` — the response is genuinely short or empty.

`recovered` is populated only on a complete page, so it never names a fault that
`degradation_reason` is already naming. Where a page is short and only part of
the fault was compensated, the whole count goes to `degradation_reason`: the
page is the unit, splitting the count across both fields would be arithmetic no
caller asked for, and it would leave the two fields disagreeing about one
request.

## Following the #608 shape rather than adding to it

`PUBLIC_DROP_CAUSE_NAMES` stays the single classification table and
`NON_DEGRADING_DROP_CAUSES` stays derived from it. No third list is introduced:
the fault set is the complement, which the code already computed inline and
which is now a named helper, `_fault_drops`, beside `_public_exclusions`. Both
response fields are built from that one table, so a cause cannot appear in both
and cannot appear in neither — asserted directly as a partition, on counts as
well as keys.

One deliberate asymmetry. `excluded` is keyed by public vocabulary; `recovered`
is keyed by the internal cause names. `archive_scope_excluded` needed
translating because the caller's word for that thing is already `archived` —
they passed it in `archive_scope`. The five fault causes have no such
counterpart: they already reach the caller under exactly these names inside
`degradation_reason`, and giving them a second public word would make one fault
answer to two names depending on whether the page happened to fill, which is the
drift this series exists to remove. Both fields use the word the caller already
knows for that thing. The field documentation says plainly that its keys are
diagnostics and that the check is emptiness, not a key.

## Tests

`backend/tests/test_search_archive_scope_vault_path_unit.py` gains seven cases,
pinning both halves and the boundary between them: a fault the refill replaced
leaves a complete page unflagged and counted in `recovered`; a fault that
exhausted the pool still degrades with `recovered` empty; a short page with no
fault behind it is not degraded; a partly-compensated fault is reported once, as
the degradation; a complete page carrying both a filter drop and a fault drop
reports each in its own field with neither borrowing the other's vocabulary; the
map is empty and present on a clean search; and `_fault_drops` is the exact
complement of `_public_exclusions`.

Mutation proof, on that file (18 cases: 11 pre-existing, 7 new). Each mutation
was applied in place, measured, and reverted:

| mutation | result | red |
| --- | --- | --- |
| baseline | `18 passed` | — |
| revert `search_service.py` to the base commit | `3 failed, 15 passed` | the two compensated-fault cases + the helper's partition |
| drop `and page_short_of_limit` from the reason | `2 failed, 16 passed` | exactly the two "complete page is not degraded" cases |
| report `recovered` unconditionally | `2 failed, 16 passed` | exactly the two short-page cases asserting it is empty |
| predicate is page length alone (drop `faults and`) | `2 failed, 16 passed` | the new "short page, no fault" guard **and** the pre-existing #604 case |
| restored | `18 passed` | — — `git status` empty |

Every pre-existing assertion in the file stays green under all four, including
the five that pin #604 and #608. The last row is the one worth noting: the naive
predicate this change could have used reopens #604, and the suite says so
without anyone having to remember it.

The OpenAPI contract assertion was observed red before `recovered` was added to
`AkbSearchEnvelope` and green after, in the same order.

Repo gate: `bash scripts/check.sh` — result reported below with counts.

## Documentation

All five surfaces move together, because the defect this series has been
unwinding came from three descriptions of one field drifting apart:

- `SearchResponse` — the `degraded` comment now states both halves of the
  condition and names the two things that never set it; `recovered` documented
  beside `excluded`.
- the OpenAPI contract — `recovered` added to `AkbSearchEnvelope`, required and
  integer-valued, matching `excluded`; pinned by the contract test.
- the checked-in generated client schema — regenerated via
  `node scripts/check-core-types.mjs --write`, not hand-edited. The client's
  type fixture had to gain the field too, which is the contract working: an
  envelope literal missing a required key stops compiling.
- the MCP tool description — the `degraded` sentence now says the page came back
  short, and tells agents not to retry or warn on `recovered`.
- `backend/CHANGELOG.md` and `docs/designs/search-filter-contract.md`.

## Not in scope

No frontend file is touched. How a surface should react to `degraded` is the
open question in #605 and is a separate decision; this change only makes the
flag mean what the surfaces already assume it means. `_hydrate_hits` and its
`dropped` dict are unchanged — same six causes, same keys, same counts, same
warning log.


imjhyuk-oss:
Local verification at `c8cfff6f`.

**Repo gate** — `bash scripts/check.sh`: `OK`. Every step ran: the 3.14 parser
preflight, the E2E suite manifest (23 curated, 34 deferred), agent-roles (8
rendered files match), ruff clean, mypy `no issues found in 385 source files`,
bandit `0 files skipped, 108143 lines scanned`, eslint (0 errors, 105
pre-existing warnings), frontend `tsc`, the `@akb/client` build / typecheck /
vitest / codegen-drift / packed-consumer proof, the markdown-editor package, and
detect-secrets.

The first run of that gate failed, usefully: `typecheck (@akb/client)` rejected
the envelope literal in `test/types.ts` for missing the new required key. The
fix is in this branch.

**Unit suite** — `pytest tests/ -k 'not _e2e'`, which is what the backend
workflow runs:

```
56 failed, 3787 passed, 326 skipped, 106 deselected in 918.64s
```

All 56 failures are in `test_git_service.py` (39), `test_external_git_runner.py`
(14) and `test_external_git_capability.py` (3), and every one of them is
`Could not resolve host: mirror.test` — the fake hostname those suites pin
through `curlopt_resolve`, which this sandbox's resolver does not provide.
Negative control on a pristine checkout of the base commit, same three files:

```
56 failed, 165 passed in 48.09s     (3 / 14 / 39, identical set)
```

Same failures, same counts, without this branch. Nothing in this change imports
or touches those modules, and no search, OpenAPI, MCP or document test is among
them. They should pass on a CI runner with working DNS.

**Focused** — `test_search_archive_scope_vault_path_unit.py` and
`test_rest_openapi_contract_unit.py`: `47 passed`. The mutation table in the PR
body was measured on the first of those (18 cases), each mutation applied in
place and reverted, ending at `18 passed` with an empty `git status`.

PR #614, opened 2026-09-18
feat(search): count search degradation and expose it on /health

### LINKED ISSUE #612: Search degradation is not counted anywhere, so its consumers cannot be chosen on evidence
## The gap

Nothing counts how often search returns `degraded`, or why.

- `/health` reports a section per queue — external git, asset GC, metadata,
  events, notifications, file projection, derived index, vector store
  (`backend/app/main.py:587-594`) — and none of them is search. The word
  `degraded` does appear there, but as the vector store's *reachability* verdict
  (`:426-428`), a different namespace from the search response field.
- `tool_usage_daily` records `(day, tool, outcome, calls, duration)`, and a
  degraded search is a successful call. It lands under `ok` alongside every
  healthy one, so the table cannot distinguish them.
- The only trace is a log line — `hydrate: dropped N hit(s): {...}` and the
  vector-store warnings — which lives as long as the pod does.

So there is no way to answer, for any deployment: how often is a search response
degraded, which cause dominates, and does the caller still receive results?

## Why those three questions specifically

They are the ones a consumer's behaviour should be chosen from, and right now it
is being chosen without them.

`degraded` means different things in different surfaces today — one renders the
results with a warning, another withholds them entirely, a third throws only when
the result set is empty. Which is right depends on how often the flag fires and
whether results survive when it does. Nobody can look that up.

The same blindness hides regressions. A cause that starts firing ten times more
often than it used to would produce no signal anywhere until somebody noticed a
banner.

## Proposal

Expose search degradation the way the queues are already exposed: a section on
`/health` with counts by cause, and a matching aggregate that survives a pod
restart.

Roughly:

```
search: {
  observed: <responses sampled>,
  degraded: <of those, how many carried the flag>,
  by_cause: {"hydration_miss": N, "sparse_encoder_degraded": N, ...},
  degraded_with_results: <how many still carried results>
}
```

The last field is the one the consumer question turns on, and it is the one no
existing surface could ever answer.

Shape notes, in the spirit of the neighbouring sections:

1. **Counts, not booleans.** Every queue section on `/health` reports numbers and
   lets the reader judge; this should match rather than invent a new verdict.
2. **By cause.** An aggregate that says only "3% degraded" cannot separate a
   write race from a retrieval outage, and those call for opposite responses.
3. **Durable enough to compare days.** In-process counters reset on every deploy,
   and deploys here are frequent; whatever backs this should outlive one.
4. **Cheap on the hot path.** Search latency already spans a wide range; the
   accounting must not add to it. Incrementing is fine, joining is not.

## Relationship to the other issues

This is the measurement that the open question in akb#605 needs — whether a fast
keyboard surface should show partial results turns on how often a degraded
response still carries usable ones, and that number does not exist yet.

It also guards the fix in the sibling issue about compensated faults: once a
complete page stops being flagged, the remaining rate is the thing worth
watching, and there would otherwise be no before-and-after to compare.

imjhyuk-oss:
Closes #612.

`/health` gains a `search` section reporting how many search responses were
observed today, how many were degraded, how many of those still carried
results, and the breakdown by cause — backed by two daily tables so the numbers
survive a deploy.

```
search: {
  day: "2026-09-18",
  observed: 41203,
  degraded: 118,
  degraded_with_results: 96,
  by_cause: {"hydration_miss": 111, "sparse_encoder_degraded": 7},
  pending_flush: 12,
  consecutive_flush_failures: 0
}
```

## The gap this closes

Three surfaces look like they should answer "how often is a search degraded",
and none can. `/health` reports a section per indexing queue and search is not a
queue — the word `degraded` does appear there, but as the vector store's
*reachability* verdict, a different namespace from the response field.
`tool_usage_daily` records `(day, tool, outcome, calls, duration)` and a
degraded search is a successful call: it lands under `ok` beside every healthy
one, so the table cannot separate them. The only remaining trace was a log line,
which lives as long as the pod does.

`degraded_with_results` is the field that matters most and the one no existing
surface could ever have produced. Consumers behave differently on the flag
today — one renders the results with a warning, another withholds them entirely
— and which is right depends on how often a degraded response still carries
usable results. That number did not exist. It does now, and it is the same
number the open question in #605 turns on.

It also gives the sibling change a before-and-after. Now that a page the refill
loop completed stops being flagged, the remaining rate is the thing worth
watching, and there would otherwise be nothing to compare it against.

## Answering the four shape constraints

**Counts, not a verdict.** The neighbouring sections report numbers and let the
reader judge, and this one has no business inventing a second opinion about
whether the service is healthy. It is deliberately *outside* the top-level
`status` aggregate: a search degraded by an ordinary write race is not queue
work left undone, and letting it move that verdict would make a delete look
like an outage. A test asserts the section carries no verdict key and that
`_aggregate_status` is unmoved by it.

**By cause.** An aggregate that says only "3% degraded" cannot separate a write
race from a retrieval outage, and those call for opposite responses.

**Durable.** `search_degradation_daily` (one row per day: the denominators) and
`search_degradation_cause_daily` (one row per day and cause: the breakdown).
Two tables rather than one because folding them together needs a sentinel cause
meaning "not degraded", and because `degraded` is *not* the sum of the cause
rows — a response whose reason names several hydration causes counts under each
— so the total is stored rather than derived.

**Cheap on the hot path.** Recording touches only memory: a dict lookup and a
few integer adds behind an already-assembled response. The database write is a
per-day *delta* flushed on a 30s timer off the request path — one small upsert
per tick, not one per response — so a busy deployment writes two rows a minute
rather than two a second. No index beyond the primary keys, no purge: the whole
cause vocabulary is a closed set, so this is single-digit rows per day, kept
indefinitely like `tool_usage_daily`.

## What is not durable, said out loud

A pod killed without a graceful stop loses up to one flush interval of counts.
That is a real limit, so rather than hiding it the section reports
`pending_flush` — exactly how much is at risk at that moment — and
`consecutive_flush_failures` beside it. A graceful stop drains before exit, so
an ordinary rolling deploy loses nothing, and a flush that fails hands its
counts back to the accumulator to be retried rather than dropping them. A
counter with an unreported hole is worse than no counter, which is also why the
per-day cause cap folds overflow into one bucket instead of discarding it.

## Two things worth review

**The chokepoint is a decorator, not a call before each `return`.** `search` has
four return sites and adding a fifth is an ordinary edit, so a chokepoint that
has to be remembered is one that will eventually be forgotten — and a
denominator with a hole in it understates every ratio built on it silently.
Wrapping also means a request that raised before producing a response is not
counted as one, which is the behaviour we want and would otherwise have had to
write out.

**The cause is parsed out of `degradation_reason`, by code that lives beside the
code that writes it.** `_hydration_drop_reason` and `degradation_causes` share
one prefix constant and are tested against each other, so the string cannot be
reshaped in one place and misread in another. An unparseable reason yields no
cause rather than a guess: a malformed string is not evidence about the corpus,
and minting a key from one would put unbounded cardinality into a counter with
a closed vocabulary.

## Auth posture

The section sits on the unauthenticated half of `/health`, with the queue
backlogs and the vector-store reachability, because it is the same class of
thing — aggregate operational counts that uptime pollers read. It carries no
query text, no vault name and no per-caller data; the cause names are the
internal vocabulary the response already publishes in `degradation_reason`. The
authenticated half (RBAC counters, audit and tool-usage internals) is unchanged.

## Tests and mutation proof

`test_search_degradation_stats_unit.py` (16 cases) holds the arithmetic against
a fake connection; `test_search_degradation_stats_postgres.py` (4 cases) holds
the SQL against a real database, including the one claim a single-writer test
can never catch — that two pods folding the same day *add* rather than
overwrite. An `ON CONFLICT` that replaced would look correct everywhere else and
silently discard a deployment's counts. The end-to-end chokepoint proof lives
beside the search harness in `test_search_archive_scope_vault_path_unit.py`:
one real search, exactly one observation, attributed to its cause.

Mutation results are in a comment below with exact counts.

Two existing suites needed updating, and both are tripwires doing their job:
`test_lifecycle_workers.py` stubs every unconditional worker start so it can
stay loop-free (it already does this for `tool_usage` and the transfer reaper,
with the same comment), and `test_access_contributions_postgres.py` pins the
migration head so a new migration cannot slip past the database its fixtures
build.


imjhyuk-oss:
Local verification at `f15d4254`.

**Repo gate** — `bash scripts/check.sh` on a clean tree at that commit: `OK`.
Every step ran: the 3.14 parser preflight, the E2E suite manifest (23 curated,
34 deferred), agent-roles (8 rendered files match), ruff clean, mypy
`no issues found in 387 source files` (385 before — the two new modules), bandit
`0 files skipped, 108501 lines scanned`, eslint (0 errors), frontend `tsc`, the
`@akb/client` build / typecheck / vitest (60) / codegen-drift / packed-consumer
proof, `@akb/markdown-editor` (53), the frontend vitest suite (145 files, 1257
tests) and detect-secrets.

**Unit suite** — `pytest tests/ -k 'not _e2e'`, which is what the backend
workflow runs:

```
56 failed, 3808 passed, 326 skipped, 106 deselected in 1029.41s
```

3787 → 3808 is exactly the 21 tests added here (16 + 4 + 1). The 56 failures are
the same pre-existing set as on the base commit — `test_git_service.py` (39),
`test_external_git_runner.py` (14), `test_external_git_capability.py` (3), every
one of them `Could not resolve host: mirror.test`, the fake hostname those
suites pin through `curlopt_resolve` and this sandbox's resolver does not
provide. Negative control against a pristine checkout of the base was run for
the sibling PR and reported the identical 56 (3 / 14 / 39). Nothing here imports
those modules.

**Mutation proof.** Suite: the two new files plus the two existing ones this
change touches — 49 cases. Each mutation applied in place, measured, reverted:

| mutation | result | red |
| --- | --- | --- |
| baseline | `49 passed` | — |
| drop the `@_counted` chokepoint from `search` | `2 failed, 47 passed` | the end-to-end observation, and the assertion that the method is wrapped |
| the daily upsert REPLACES instead of accumulating | `1 failed, 48 passed` | the two-pods-fold-the-same-day proof, against a real database |
| a failed flush drops its counts instead of giving them back | `1 failed, 48 passed` | the give-back case |
| the snapshot ignores the un-flushed delta | `1 failed, 48 passed` | the delta-fold case |
| only degraded responses count toward `observed` | `4 failed, 45 passed` | the denominator, and the three cases that read it |
| cause overflow is dropped instead of folded | `1 failed, 48 passed` | the fold case |
| the parser never splits a hydration reason | `6 failed, 43 passed` | the formatter/parser round trip, the multi-cause split, and four downstream |
| the section leaks a queue key into the top-level aggregate | `1 failed, 48 passed` | the no-verdict case |
| restored | `49 passed` | — — `git status` empty |

The second row is the one to read: an `ON CONFLICT` that overwrote instead of
adding passes every single-writer test and silently discards one deployment's
counts, so the only thing that catches it is a real database being written
twice for the same day.

PR #637, opened 2026-09-21
fix(graph): resolve document endpoints against the active authority

skt8776-oss:
## What broke

On a `postgres_native` installation, **every document created after the vault's cutover was invisible to the graph layer**:

- `akb_link` refused it as a source or target — `{"code": "not_found", "message": "Source resource not found: akb://…"}` — while `akb_get`, `akb_browse` and `akb_search` served the same URI normally.
- The implicit `links_to` edges that `_store_edge` extracts from a body referencing it were dropped without a sign.

Pre-cutover documents linked fine, so the failure looked path- or collection-specific. It is authority-specific.

## Why

`kg_service._resource_exists` resolved a document endpoint against the legacy `documents` catalog alone. The native arm does not write that catalog — `document_service.newest_public_slug` says so in as many words: *"the native-ledger arm keeps no legacy projection"*. Same family as akb#525, which fixed the counting and freshness surfaces; this is the identity surface the graph needs.

## Fix

`_resource_exists` now asks the native ledger too, through the selector `document_counters.native_documents_are_authoritative()` (the one search hydration already uses, so the graph cannot disagree with search) and the same population — `surface = 'document'`, `lifecycle = 'live'`, which counts an archived document and excludes a deleted one.

Both arms are consulted on a native installation, not just the authoritative one: a cutover leaves pre-cutover documents in the catalog, and an endpoint that resolved yesterday must keep resolving. A legacy installation never touches the native tables.

Tables and files are unchanged — files are projected into `vault_files` by the native file projection, so their check still answers.

## Reproduced against a live native instance

| step | result |
|---|---|
| link two pre-cutover documents | `200 {"linked": true}` |
| link from a document created minutes earlier (new collection) | `404 not_found` |
| same, in a long-established collection | `404 not_found` |
| `akb_get` / `akb_browse` / `akb_relations` on that same URI | resolve normally |

(The probe documents and the diagnostic edge were removed afterwards.)

## Tests

`backend/tests/test_kg_vault_boundary_unit.py` gains four cases: a native-only document is a valid endpoint; a pre-cutover document still resolves on a native installation; a legacy installation never queries the native ledger; a document missing from both arms is rejected. The first and fourth fail on `main` and pass here.

`pytest backend/tests --ignore=tests/mcp_e2e --ignore=tests/concurrency --ignore=tests/eval --ignore=tests/bench` → **2955 passed, 643 skipped**. The endpoint-driven e2e suites were not run in this environment (they need a runtime descriptor).

## Known gaps left open (not this PR)

1. **Implicit edges are still not written on the native arm.** `store_document_relations` is called only from `document_service`; `native_document_service` validates structured relation refs but never stores edges. So a native document's body links and its `depends_on` / `related_to` produce no edges even after this fix — explicit `akb_link` is what starts working here.
2. **`_batch_resolve_names` reads `documents` for titles**, so a native endpoint appears in `akb_relations` without a `name`. Degrades gracefully; resolving it means reading the head revision's frontmatter.
3. **A move does not register its collections**, so a document moved into a new path is served but its folders do not appear in a parent `akb_browse` (`akb_create_collection` after the move is the workaround). Same legacy-projection family.

Happy to take any of these next if you'd like them in one go.

https://claude.ai/code/session_01Atxss6Kf2hMqr7FbWDLngm


imjhyuk-oss:
## Follow-up verification: native graph parity

This PR fixes native canonical-URI endpoint checks. #656 identifies the non-URI lookup path that remains outside that fix. The subsequent lifecycle changes in #638 also need the corrections tracked in #654 and #655.

Measured follow-up blockers are now tracked together in #654, #655, and #656. Each issue includes a runnable isolated PostgreSQL fixture, exact revisions, before/after observations, measurement limits, and release-unblock criteria.

The comparison is #638's parent `f2cc7af2` (already includes #637) versus main `1d0e023a`; it is a real PostgreSQL service/SQL spike, not an HTTP/MCP end-to-end test or a claim about the deployed artifact.

| Follow-up | Measured result | Classification |
| --- | --- | --- |
| #654 | After delete/recreate at the same path, an old delete intent leaves 1 replacement link before #638, but 0 on main. | New destructive regression in #638. |
| #655 | Ordered A→B→C moves improve from edge A to edge C; reordered hooks leave edge A before #638 and edge B on main, even after the worker runs. | Incomplete repair of an existing move defect. |
| #656 | Ordinary Markdown path and wiki references store 0 edges on both revisions; canonical URI references store 1. | Existing non-URI resolution defect remains. |

Please track these as release blockers for native graph-parity acceptance. The three issue bodies cross-reference both original PRs and each other. These links do not mark either merged PR as a fix for the new issues, and do not claim that every finding is a new regression.

PR #638, opened 2026-09-21
fix(native): keep the collections catalog and the graph in step with native writes

skt8776-oss:
Follow-up to #637, which fixed one of four surfaces that read legacy rows and go blank for a native document. This PR closes the other three.

The native arm writes no legacy `documents` row — that is the design (`document_service.newest_public_slug`: *"the native-ledger arm keeps no legacy projection"*). What is not the design is browse losing folders and the graph losing edges.

## 1. Folders vanished from browse

`collections` rows are what browse renders a folder from, and only the legacy write path maintained them. Observed on a live native instance: a document moved into `direction/skh-project/architecture/` was served at its path and counted in the collection totals, while `akb_browse` on `direction` showed no such folder — and `akb_delete_collection` on it answered `not_found`, because no row had ever existed.

`put` and `move` now call the same `CollectionRepository.get_or_create` the legacy arm calls, for the document's own collection (not its ancestors — the legacy arm does not either).

Best-effort by design: the Revision is already committed and is the authority. A catalog row that could not be written is a browse defect, not a lost write, and the next document into the same collection retries it. Failing here would report a successful write as an error.

## 2. A move dropped the document's edges

The legacy move rewrites every edge naming the old URI (both endpoint columns, with the duplicate guard). The native move did not — so after #637 made `akb_link` work, the first move would have silently orphaned those links.

The rewrite is now one shared helper, `kg_service.relink_resource_edges`, called by both arms (first commit is that move, no behaviour change).

## 3. Body links produced no edges at all

`store_document_relations` is called only from the legacy write path, so on the native arm a document's markdown links and its `depends_on` / `related_to` / `implements` never became edges. The graph for a post-cutover vault was empty except for whatever `akb_link` put in it.

Placed in the derived rewrite rather than the write path: the derived worker already re-reads the verified Head, locks the resource, and replaces derived state in one transaction — the same place chunks are rebuilt, with the same supersession check. It now also:

- rewrites the document's implicit edges from the verified body,
- clears the **previous** path's implicit rows when the path changed (`store_document_relations` only clears the URI it is writing), and
- drops every edge of a deleted document, matching the legacy delete.

Explicit `akb_link` rows are never touched by the derived path; the move carries them.

## Correction to #637's "known gaps"

That PR listed "`_batch_resolve_names` reads `documents`, so a native endpoint appears without a `name`". That is wrong: the helper already falls back to the document's path for anything the catalog cannot name, so native endpoints render with their path. What is missing is only the human **title**, and resolving it means reading and verifying the head payload the way search hydration does. Duplicating that payload join into the graph module is not worth a display label, so it stays open deliberately — if it is wanted, the right shape is one shared title resolver next to search's hydration, not a second copy of the join.

## Tests

`backend/tests/test_native_derived_worker_unit.py` (+6): the relation parse (frontmatter lists, and a scalar where a list belongs), the live rewrite storing body links, a moved rewrite clearing the old path's implicit rows, a superseded revision writing nothing, delete clearing a document's edges, delete of a *file* leaving the document graph alone.

`backend/tests/test_native_legacy_catalog_parity_unit.py` (new, 5): a native write registers its collection, a vault-root document registers nothing, a catalog failure logs and does not fail the committed write, a move relinks both URIs, a no-op move touches nothing.

Every new derived-worker test fails on `main` (one assertion failure, five errors on the missing import path).

`pytest backend/tests --ignore=tests/mcp_e2e --ignore=tests/concurrency --ignore=tests/eval --ignore=tests/bench` → **3530 passed, 824 skipped**, plus 2 pre-existing failures in `test_seahorse_cloud_mgmt_url_unit.py` that reproduce with these changes stashed (environment-dependent, unrelated). The endpoint-driven e2e suites were not run locally (they need a runtime descriptor); CI runs them here.

https://claude.ai/code/session_01Atxss6Kf2hMqr7FbWDLngm


imjhyuk-oss:
## Follow-up verification: native graph parity

Collection registration and ordinary move handling are improved by this PR. The new asynchronous graph deletion has a replacement-resource regression (#654), move hooks still lack ordering guarantees (#655), and the newly used relation writer retains legacy-only non-URI resolution (#656). Related endpoint-check change: #637.

Measured follow-up blockers are now tracked together in #654, #655, and #656. Each issue includes a runnable isolated PostgreSQL fixture, exact revisions, before/after observations, measurement limits, and release-unblock criteria.

The comparison is #638's parent `f2cc7af2` (already includes #637) versus main `1d0e023a`; it is a real PostgreSQL service/SQL spike, not an HTTP/MCP end-to-end test or a claim about the deployed artifact.

| Follow-up | Measured result | Classification |
| --- | --- | --- |
| #654 | After delete/recreate at the same path, an old delete intent leaves 1 replacement link before #638, but 0 on main. | New destructive regression in #638. |
| #655 | Ordered A→B→C moves improve from edge A to edge C; reordered hooks leave edge A before #638 and edge B on main, even after the worker runs. | Incomplete repair of an existing move defect. |
| #656 | Ordinary Markdown path and wiki references store 0 edges on both revisions; canonical URI references store 1. | Existing non-URI resolution defect remains. |

Please track these as release blockers for native graph-parity acceptance. The three issue bodies cross-reference both original PRs and each other. These links do not mark either merged PR as a fix for the new issues, and do not claim that every finding is a new regression.

